Issue #10, opened 2018-11-15
Support installing a named version


ry:
EG `curl -sSf https://raw.githubusercontent.com/denoland/deno_install/master/install.py v0.1.12 | python`

MarkTiedemann:
@ry I think this can be closed.

ry:
Thx

PR #11, opened 2018-11-16
Install specific version of deno

kt3k:
With this change, we can install the specified version of deno with the following command:

```
curl -sSf https://raw.githubusercontent.com/denoland/deno_install/master/install.py | python - v0.1.12
```

refs: #10


kt3k:
Updated documentation!

PR #17, opened 2018-12-30
Update install.py to add .exe for windows

matt1:
Update install.py to use '.exe' extension on windows so that `deno` can run successfully.  Addresses issue https://github.com/denoland/deno/issues/1429

### DIFF
--- a/install.py
+++ b/install.py
@@ -81,7 +81,7 @@ def download_with_progress(url):
 
 def main():
     bin_dir = deno_bin_dir()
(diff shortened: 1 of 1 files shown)


hayd:
Did this patch not work?

matt1:
Worked, but tests failed.   Figured that I don't know enough about _why_ it is deno vs deno.exe - i.e. what was the rationale for originally not using .exe for windows in the first place.  So instead of changing the tests for something I have minimal knowledge about, I've abandoned this PR instead.

PR #18, opened 2019-01-04
Support installing a named version with powershell script

artemmnad:
Add support for specifying the version of deno when installing it with `install.ps1`.

```powershell
$install = iwr https://deno.land/x/install/install.ps1; iex "$install v0.2.0"
```

If no version is specified, the latest version is installed.

Closes: #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,4 +30,8 @@ If you need to install specific version of deno, use the following commands:
 curl -L https://deno.land/x/install/install.py | python - v0.2.0
 ```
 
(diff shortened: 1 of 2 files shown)


ry:
I'm going to go with #19 instead because @srijanreddy98 has also got it working in deno_std's CI https://github.com/denoland/deno_std/pull/89
If you don't mind, have a look at the code and to see if there's any improvements that can be made.
Thanks!

PR #21, opened 2019-01-05
Add deno-install

### LINKED ISSUE #14: Add installer written in deno
If you have a prior version of Deno installed, it would be cool to upgrade using this:
```
> deno https://deno.land/x/install/install.ts 
```

chiefbiiko:
Adds an installer written in deno.

resolves #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,14 @@ iex (iwr https://deno.land/x/install/install.ps1)
 
 _Note: Depending on your security settings, you may have to run `Set-ExecutionPolicy RemoteSigned -Scope CurrentUser` first to allow downloaded scripts to be executed._
 
+**Update with deno:**
+
+```
+deno --allow-write --allow-net --allow-env --allow-run https://deno.land/x/install/install.ts
+```
+
+_Note: Installs the latest `deno` version by default._
+
 ## Install other versions
 
 If you need to install specific version of deno, use the following commands:
@@ -34,4 +42,10 @@ curl -L https://deno.land/x/install/install.py | python - v0.2.0
 
 ```
 iwr https://deno.land/x/install/install.ps1 -out install.ps1; .\install.ps1 v0.2.0
+```
+
+**(Re)install with deno:**
+
+```
+deno --allow-write --allow-net --allow-env --allow-run https://deno.land/x/install/install.ts v0.2.0
 ```
--- a/install.ts
+++ b/install.ts
@@ -1,0 +1,218 @@
+import {
+  Process,
+  args,
+  chmod,
+  copyFile,
+  env,
+  exit,
+  platform,
+  writeFile,
+  removeAll,
+  rename,
+  run,
+  makeTempDir
+} from "deno";
+
+import { mkdirp } from "https://deno.land/x/std/fs/mkdirp.ts";
+import { join, resolve } from "https://deno.land/x/std/fs/path/mod.ts";
+
+const MAX_FOLLOWS: number = 4;
+const isWindows: boolean = platform.os === "win";
+const procEnv: { [key: string]: any } = env();
+const PATH: string = procEnv.PATH || procEnv.Path;
+
+function getHome(): string {
+  return isWindows
+    ? resolve(procEnv.HOMEDRIVE, procEnv.HOMEPATH)
+    : procEnv.HOME;
+}
+
+const RELEASE_URL: string =
+  "https://api.github.com/repos/denoland/deno/releases";
+const LATEST_RELEASE_URL: string = `${RELEASE_URL}/latest`;
+const TAG_RELEASE_URL: string = `${RELEASE_URL}/tags`;
+
+const DENO_DIR: string = join(getHome(), ".deno");
+const DENO_BIN_DIR: string = join(DENO_DIR, "bin");
+const DENO_BIN: string = join(DENO_BIN_DIR, isWindows ? "deno.exe" : "deno");
+const OLD_DENO_BIN: string = DENO_BIN.replace(/deno(\.exe)?$/, "old_deno$1");
+const LINUX_GZIP: string = "deno_linux_x64.gz";
+const OSX_GZIP: string = "deno_osx_x64.gz";
+const WIN_ZIP: string = "deno_win_x64.zip";
+
+function panic(err?: Error): void {
+  if (err) {
+    console.error("[deno-install error]", err.stack);
+  }
+  console.error("[deno-install error]", "Installation failed");
+  exit(1);
+}
+
+function pinup(...args: any): void {
+  console.log("[deno-install info]", ...args);
+}
+
+async function runBufferStdout(args: string[]): Promise<string> {
+  const p = run({ args, stdout: "piped" });
+  const stdout = await p.output();
+  if (!(await p.status()).success) {
+    panic(Error(`Running ${args} failed`));
+  }
+  p.close();
+  return new TextDecoder().decode(stdout);
+}
+
+async function follow(url: string): Promise<any> {
+  let res: any; // TODO: annotate deno Response
+  let located: boolean = false;
+  let count: number = 0;
+  while (!located) {
+    res = await fetch(url, { headers: { "User-Agent": "deno" } });
+    if (res.status >= 300 && res.status < 400) {
+      if (++count >= MAX_FOLLOWS) {
+        panic(Error(`Unable to fetch from ${url}`));
+      }
+      url = res.headers.get("Location");
+    } else if (res.status === 200) {
+      located = true;
+    } else {
+      panic(Error(`HTTP error ${res.status} ${res.statusText}`));
+    }
+  }
+  return res;
+}
+
+async function releaseUrl(tag?: string): Promise<{ [key: string]: string }> {
+  const url: string = tag ? `${TAG_RELEASE_URL}/${tag}` : LATEST_RELEASE_URL;
+  let filename: string;
+  switch (platform.os) {
+    case "linux":
+      filename = LINUX_GZIP;
+      break;
+    case "mac":
+      filename = OSX_GZIP;
+      break;
+    case "win":
+      filename = WIN_ZIP;
+      break;
+    default:
+      panic(Error(`Unsupported operating system ${platform.os}`));
+  }
+  const res: any = await follow(url); // TODO: annotate deno Response
+  const release: { [key: string]: any } = await res.json();
+  const archive: { [key: string]: any } = release.assets.find(
+    (asset: { [key: string]: any }) => asset.name === filename
+  );
+  if (!archive || !archive.browser_download_url) {
+    panic(Error(`Unable to find ${filename} @ ${url}`));
+  }
+  return {
+    url: archive.browser_download_url,
+    tag: archive.browser_download_url.replace(/^.*(v\d+\.\d+\.\d+).*$/, "$1")
+  };
+}
+
+async function tempDownload(
+  tempDir: string,
+  url: string,
+  suffix: string
+): Promise<string> {
+  const res: any = await follow(url); // TODO: annotate deno Response
+  const tempFile: string = join(tempDir, `${Date.now()}.${suffix}`);
+  await writeFile(tempFile, new Uint8Array(await res.arrayBuffer()));
+  return tempFile;
+}
+
+async function unpackBin(archive: string): Promise<void> {
+  await mkdirp(DENO_BIN_DIR);
+  let args: string[];
+  if (isWindows) {
+    await rename(DENO_BIN, OLD_DENO_BIN);
+    args = [
+      "powershell.exe",
+      "-Command",
+      `Expand-Archive "${archive}" -DestinationPath "${DENO_BIN_DIR}"`
+    ];
+  } else {
+    args = ["gunzip", "-d", archive];
+  }
+  const p: Process = run({ args });
+  if (!(await p.status()).success) {
+    panic(Error(`Running ${args} failed`));
+  }
+  p.close();
+  if (!isWindows) {
+    if (platform.os === "linux") {
+      await rename(DENO_BIN, OLD_DENO_BIN);
+    }
+    await copyFile(archive.replace(/\.gz$/, ""), DENO_BIN);
+  }
+}
+
+async function makeHandy(): Promise<void> {
+  if (isWindows) {
+    if (!PATH.toLocaleLowerCase().includes(DENO_BIN_DIR.toLocaleLowerCase())) {
+      let updatedPath: string = PATH;
+      if (PATH.endsWith(";")) {
+        updatedPath += DENO_BIN_DIR;
+      } else {
+        updatedPath += `;${DENO_BIN_DIR}`;
+      }
+      const args: string[] = [
+        "powershell.exe",
+        "-Command",
+        `[Environment]::SetEnvironmentVariable("PATH","${updatedPath}",` +
+          `[EnvironmentVariableTarget]::User)`
+      ];
+      const p: Process = run({ args });
+      if (!(await p.status()).success) {
+        panic(Error(`Running ${args} failed`));
+      }
+      p.close();
+    }
+  } else {
+    await chmod(DENO_DIR, 0o744);
+    await chmod(DENO_BIN, 0o744);
+    if (!PATH.includes(DENO_BIN_DIR)) {
+      pinup(
+        `Now manually add ${DENO_BIN_DIR} to your $PATH.\nExample:\n\t\t` +
+          `echo "export PATH=${DENO_BIN_DIR}:$PATH" >> $HOME/.bash_profile`
+      );
+    }
+  }
+}
+
+async function checkVersion(tag: string): Promise<void> {
+  const output: string = await runBufferStdout(["deno", "--version"]);
+  if (!RegExp(tag.replace(/^v/, "").replace(/\./g, "\\.")).test(output)) {
+    panic(Error("Version mismatch"));
+  }
+}
+
+async function main(): Promise<void> {
+  let tag: string;
+  // if we got a truthy first argument that is a version use it as tag
+  // else if we got a truthy first argument that is not a version panic
+  // else tag remains undefined and releaseUrl(tag?) falls back to latest
+  if (/^v\d+\.\d+\.\d+$/.test(args[1])) {
+    tag = args[1];
+  } else if (args[1]) {
+    panic(Error("Malformatted tag. Examples: v0.2.7, v0.3.0"));
+  }
+  pinup(tag ? `Installing deno ${tag}` : "Installing deno@latest");
+  const actual: { [key: string]: string } = await releaseUrl(tag);
+  const tempDir: string = await makeTempDir();
+  pinup(`Downloading ${actual.url}`);
+  const tempFile: string = await tempDownload(
+    tempDir,
+    actual.url,
+    isWindows ? "zip" : "gz"
+  );
+  await unpackBin(tempFile);
+  await makeHandy();
+  await checkVersion(actual.tag);
+  await removeAll(tempDir);
+  pinup(`Successfully installed deno ${actual.tag}`);
+}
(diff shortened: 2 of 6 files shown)


install.ts
```diff
@@ -0,0 +1,203 @@
...
+  dest_dir: string
+): Promise<string[]> {
+  const script_file: string = join(temp_dir, "wunzip.vbs");
+  const script_data: Uint8Array = new TextEncoder().encode(
+    `
+    set objShell = CreateObject("Shell.Application")
+    set FilesInZip=objShell.NameSpace("${zip_file}").items
+    objShell.NameSpace("${dest_dir}").CopyHere(FilesInZip)
+    Set objShell = Nothing
+  `
```
MarkTiedemann: Instead of doing this with Visual Basic Script, how about using PowerShell for the unzipping? With PowerShell, it would be `Expand-Archive C:\file.zip -DestinationPath C:\directory`, and you wouldn't need to write a temporary script file, but instead pass the command to PowerShell via `powershell.exe -Command {CommandTextHere}`.

chiefbiiko: @MarkTiedemann thanks, that is way less clutter! switched to ps unzipping.

install.ts
```diff
@@ -0,0 +1,185 @@
...
+
+import { mkdirp } from "https://deno.land/x/std/mkdirp/mkdirp.ts";
+import { join, win32 } from "https://deno.land/x/path/index.ts";
+
+const WIN32: boolean = platform.os === "win";
+
+const proc_env: { [key: string]: any } = env();
+
+function get_home(): string {
+  return WIN32 ? win32.resolve("C:", proc_env.HOMEPATH) : proc_env.HOME;
```
MarkTiedemann: Instead of hard-coding `"C:"`, you could use `HOMEDRIVE`.

```suggestion
  return WIN32 ? win32.resolve(proc_env.HOMEDRIVE, proc_env.HOMEPATH) : proc_env.HOME;
```

install.ts
```diff
@@ -0,0 +1,185 @@
...
+const DENO_REPO_URL: string = "https://github.com/denoland/deno";
+const LATEST_RELEASE_URL: string = `${DENO_REPO_URL}/releases/latest`;
+const TAG_RELEASE_URL: string = `${DENO_REPO_URL}/releases/tag`;
+
+const DENO_DIR: string = join(get_home(), ".deno");
+const DENO_BIN_DIR: string = join(DENO_DIR, "bin");
+const DENO_BIN: string = join(DENO_BIN_DIR, WIN32 ? "deno.exe" : "deno");
+const OLD_DENO_BIN: string = DENO_BIN.replace(/deno(\.exe)?$/, "old_deno$1");
+const DENO_LINK: string = join(
+  WIN32 ? win32.resolve("C:", "Windows", "System32") : "/usr/local/bin",
```
MarkTiedemann: You could use the `SystemRoot` environment variable here (e.g. `SystemRoot=C:\WINDOWS`).

```suggestion
  WIN32 ? win32.resolve(proc_env.SystemRoot, "System32") : "/usr/local/bin",
```

install.ts
```diff
@@ -0,0 +1,185 @@
...
+    panic(Error(`(g)unzip failed with code ${child_status.code}`));
+  if (!WIN32) {
+    const gunzipd: string = archive.replace(/\.gz$/, "");
+    if (platform.os === "linux") await rename(DENO_BIN, OLD_DENO_BIN);
+    await copyFile(gunzipd, DENO_BIN);
+  }
+  child.close();
+}
+
+async function make_handy(): Promise<void> {
```
MarkTiedemann: This `make_handy` function is not consistent with the other installers (which, too, are not consistent). The Python installer prints a message that the user should add `deno` to the `PATH` manually. The PowerShell installer directly adds `deno` to the `PATH` (since it's kind of easy to do with PowerShell).

I don't think `chmod` is needed on Windows. And I don't think `symlink`ing to System32 is common practice for installers on Windows either.

@chiefbiiko What's your opinion on how to proceed with this?

chiefbiiko: IMHO ideally the installers should not require any manual steps
I found cross-platform path manipulation a little hard to implement, and ended up `symlink`ing `/usr/local/bin/deno` on linux and mac, which works just fine.
For windows path manipulation with powershell seems best to me.
Reflected this in the updated `make_handy` function.

install.ts
```diff
@@ -0,0 +1,202 @@
...
+  if (WIN32) {
+    if (!proc_env.Path.includes(DENO_BIN_DIR)) {
+      const upd_path: string = `${proc_env.Path};${DENO_BIN_DIR}`;
+      const args: string[] = [
+        "powershell.exe",
+        "-Command",
+        `[Environment]::SetEnvironmentVariable("PATH","${upd_path}",` +
+          `[EnvironmentVariableTarget]::User)`
+      ];
```
MarkTiedemann: In Windows, there's two ways to modify env vars: `setx` modifies an env var for all future terminal sessions but doesn't change the value for the current session. `set` modifies an env var for the current terminal session but doesn't persist it for future sessions.

The PowerShell command above is basically only running `setx`, setting the value for future sessions. So `deno` won't be available on the Path of the current terminal session, and users would have to restart the terminal first.

This is why the `install.ps1` script is running `$env:PATH += ";${deno_dir}"` after `[Environment]::SetEnvironmentVariable`, by the way. However, since PowerShell is executed as a (child?) process here, I don't think that's gonna work (but I can't test it since I'm not on a Windows machine at the moment).

@chiefbiiko Can you look into this perhaps? :)

install_test.sh
```diff
@@ -0,0 +1,11 @@
+if [[ $TRAVIS_OS_NAME == "linux" ]]; then 
+  ./install.py
+  ./install_test.py
+  sudo ln -s ~/.deno/bin/deno /usr/local/bin/deno
+else
+  curl https://raw.githubusercontent.com/chiefbiiko/deno-bash-installer/master/install.sh | sudo bash
+fi
+
+sudo deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
```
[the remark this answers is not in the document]
chiefbiiko: Omitting `sudo` leads to a PermissionDenied error:

```
$ deno --allow-write --allow-net --allow-env --allow-run ./install.ts
Compiling /home/travis/build/chiefbiiko/deno-install/install.ts
Downloading https://deno.land/x/std/mkdirp/mkdirp.ts...
PermissionDenied: Permission denied (os error 13)
    at DenoError (deno/js/errors.ts:19:5)
    at maybeError (deno/js/errors.ts:38:12)
    at maybeThrowError (deno/js/errors.ts:26:15)
    at sendSync (deno/js/dispatch.ts:67:5)
    at codeFetch (deno/js/os.ts:46:19)
    at resolveModule (deno/js/compiler.ts:535:38)
    at moduleNames.map.name (deno/js/compiler.ts:686:31)
    at resolveModuleNames (deno/js/compiler.ts:678:24)
    at compilerHost.resolveModuleNames (deno/third_party/node_modules/typescript/lib/typescript.js:32750:71)
    at resolveModuleNamesWorker (deno/third_party/node_modules/typescript/lib/typescript.js:20688:69)
The command "deno --allow-write --allow-net --allow-env --allow-run ./install.ts" exited with 1.
```

install.ts
chiefbiiko: I looked into possible options but could not make it work without any manual steps.
However, whenever `./install.ts` is run, deno is already installed, and the `PATH` should be set accordingly.
In the case that the binary is in place but the `PATH` is not set, `install.ts` takes care of updating the `PATH` (manual shell restart required).
As the latter should be rather unusual, informing the user about a required shell restart is okay in my opinion.

@MarkTiedemann What do you think?

sh7dm:
@chiefbiiko thanks, camelCase looks much better than snake_case

install.ts
MarkTiedemann: Haha, true, for a second there I forgot we're installing deno from an existing installation! I think it's fine as it is then. :)

MarkTiedemann: I even think the "start a fresh shell session" message is unnecessary in this case.

install.ts
```diff
@@ -0,0 +1,202 @@
...
+  if (!WIN32) {
+    if (platform.os === "linux") await rename(DENO_BIN, OLD_DENO_BIN);
+    await copyFile(archive.replace(/\.gz$/, ""), DENO_BIN);
+  }
+}
+
+async function makeHandy(): Promise<void> {
+  if (WIN32) {
+    if (!procEnv.Path.includes(DENO_BIN_DIR)) {
+      const upd_path: string = `${procEnv.Path};${DENO_BIN_DIR}`;
```
MarkTiedemann: nit

```suggestion
const updatedPath: string = `${procEnv.Path};${DENO_BIN_DIR}`;
```

Also, if you really want to cover all edge cases, you may want to check whether you actually need the semicolon. If the path is an empty string, you are unnecessarily adding a leading one. If the path already ends with a semicolon, you are unnecessarily adding a second one. Maybe splitting by semicolon, then joining with semicolon is a good idea here?

install.ts
```diff
@@ -0,0 +1,202 @@
...
+  child.close();
+  if (!WIN32) {
+    if (platform.os === "linux") await rename(DENO_BIN, OLD_DENO_BIN);
+    await copyFile(archive.replace(/\.gz$/, ""), DENO_BIN);
+  }
+}
+
+async function makeHandy(): Promise<void> {
+  if (WIN32) {
+    if (!procEnv.Path.includes(DENO_BIN_DIR)) {
```
MarkTiedemann: Windows file paths are case-insensitive, `String#includes` is not.

```suggestion
    if (!procEnv.Path.toLocaleLowerCase().includes(DENO_BIN_DIR.toLocaleLowerCase())) {
```

MarkTiedemann:
2 more nits, otherwise the win32 parts look good to me.

install_test.sh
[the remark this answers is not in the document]
hayd: Perhaps because you previously used sudo to write to the file?

ry:
I'll await approval from @MarkTiedemann on the win32 parts.

install.ts
```diff
@@ -0,0 +1,205 @@
...
+  child.close();
+  if (!WIN32) {
+    if (platform.os === "linux") await rename(DENO_BIN, OLD_DENO_BIN);
+    await copyFile(archive.replace(/\.gz$/, ""), DENO_BIN);
+  }
+}
+
+async function makeHandy(): Promise<void> {
+  if (WIN32) {
+    if (!procEnv.Path.toLocaleLowerCase().includes(DENO_BIN_DIR)) {
```
MarkTiedemann: On my Windows machine, HOMEDRIVE and HOMEPATH are both uppercase. Since they are included in DENO_BIN_DIR, a single toLocaleLowerCase is not enough in this case. Both DENO_BIN_DIR and the PATH need to be converted to either lower- or uppercase before comparing.

install.ts
```diff
@@ -0,0 +1,205 @@
...
+      const ps: Process = run({ args });
+      const psStatus: ProcessStatus = await ps.status();
+      if (!psStatus.success)
+        panic(Error(`Unable to edit PATH. ${args} -> ${psStatus.code}`));
+      ps.close();
+    }
+  } else {
+    await chmod(DENO_DIR, 0o744);
+    await chmod(DENO_BIN, 0o744);
+    if (!procEnv.Path.toLocaleLowerCase().includes(DENO_BIN_DIR)) {
```
MarkTiedemann: This else-block is for non-win32 systems. And I don't think file paths are case-insensitive there, so I think the toLocaleLowerCase needs to be removed here.

MarkTiedemann:
Still 3 minor changes needed (could be done later, too). Otherwise, LGTM.

install_test.sh
```diff
@@ -0,0 +1,11 @@
+if [[ $TRAVIS_OS_NAME == "linux" ]]; then 
+  ./install.py
+  ./install_test.py
+  sudo ln -s ~/.deno/bin/deno /usr/local/bin/deno
```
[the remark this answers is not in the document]
ry: Please remove this line

install.ts
```diff
@@ -0,0 +1,205 @@
...
+    stdout: "piped"
+  });
+  const denoStatus: ProcessStatus = await denoProc.status();
+  if (!denoStatus.success) panic(Error("Test run failed"));
+  const denoStdout: Uint8Array = new Uint8Array(32);
+  while ((await denoProc.stdout.read(denoStdout)).nread < 16);
+  denoProc.close();
+  const output: string = new TextDecoder().decode(denoStdout);
+  if (!RegExp(tag.replace(/^v/, "").replace(/\./g, "\\.")).test(output))
+    panic(Error("Version mismatch"));
```
ry: Run prettier - the formatting here seems non-standard

install_test.sh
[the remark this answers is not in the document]
ry: Yeah maybe use `$HOME/.deno/bin/deno`... although it seems odd that it would keep the symlink in `/usr/local/bin`

install_test.sh
```diff
@@ -0,0 +1,11 @@
+if [[ $TRAVIS_OS_NAME == "linux" ]]; then
```
ry: Add `set -e`

install_test.sh
```diff
@@ -0,0 +1,11 @@
...
+  ./install.py
+  ./install_test.py
+  sudo ln -s ~/.deno/bin/deno /usr/local/bin/deno
+else
+  curl https://raw.githubusercontent.com/chiefbiiko/deno-bash-installer/master/install.sh | sudo bash
+fi
+
+sudo deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
+
+deno --help
```
ry: Test the version like this
```
$(deno -v) | grep v0.2.0
```
Along with the above `set -e` this should suffice

install_test.sh
```diff
@@ -1,11 +1,19 @@
...
 fi
 
-sudo deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
+if [[ -z $NOTPATH ]]; then
+  echo PATH=$PATH:$HOME/.deno/bin >> $HOME/.bash_profile
+  echo export PATH >> $HOME/.bash_profile
+  source $HOME/.bash_profile
+  sudo deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
+else
+  sudo $HOME/.deno/bin/deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
```
ry: remove sudos

install.ts
```diff
@@ -57,20 +54,20 @@ function pinup(...args: any): void {
...
   let res: any; // TODO: annotate deno Response
+  let located: boolean = false;
   let count: number = 0;
   while (!located) {
-    if (++count === 4) {
-      panic(Error(`Unable to fetch from ${url}`));
-    }
     res = await fetch(url);
     if (res.status >= 300 && res.status < 400) {
+      if (++count === 4) {
```
hayd: >= may be a safer choice here.

Should 4 be named? E.g. MAX_FOLLOW_DEPTH, MAX_FOLLOWS or something?

hayd: `>=`

chiefbiiko: yea, the 4 is too magic

install_test.sh
chiefbiiko: just switched to `set -eo pipefail` to make sure we capture pipeline errors

chiefbiiko:
from [one of the latest test jobs](https://travis-ci.com/denoland/deno_install/jobs/170033612):

```
+deno --allow-write --allow-net --allow-env --allow-run ./install.ts v0.2.0
Unrecognized option: 'allow-run'
The command "bash -x ./install_test.sh" exited with 1.
```

What's wrong with the options?

install.ts
```diff
@@ -0,0 +1,234 @@
...
+    args: ["deno", "--version"],
+    stdout: "piped"
+  });
+  const denoStatus: ProcessStatus = await denoProc.status();
+  if (!denoStatus.success) {
+    panic(Error("Test run failed"));
+  }
+  const denoStdout: Uint8Array = new Uint8Array(32);
+  while ((await denoProc.stdout.read(denoStdout)).nread < 16);
+  denoProc.close();
```
ry: This is not a robust way to do buffer the stdout. Use this utility function (untested):

```ts
async function runBufferStdout(args: string[]): Promise<string> {
  const p = run(args, stdout: "piped");
  const stdout = await p.output();
  p.close();
  return new TextDecoder().decode(stdout);
}
```

install.ts
```diff
@@ -0,0 +1,234 @@
...
+  const output: string = new TextDecoder().decode(denoStdout);
+  if (!RegExp(tag.replace(/^v/, "").replace(/\./g, "\\.")).test(output)) {
+    panic(Error("Version mismatch"));
+  }
+}
+
+async function main(): Promise<void> {
+  let tag: string;
+  if (/^v\d+\.\d+\.\d+$/.test(args[1])) {
```
ry: Please add a comment about what you're doing here. If args[1] exists but does not look like a version, you should print an error message and exit

install.ts
```diff
@@ -0,0 +1,234 @@
...
+  const actual: { [key: string]: string } = await releaseUrl(tag);
+  const tempDir: string = await makeTempDir();
+  pinup(`Downloading ${actual.url}`);
+  const tempFile: string = await tempDownload(
+    tempDir,
+    actual.url,
+    isWindows ? "zip" : "gz"
+  );
+  await unpackBin(tempFile);
+  await makeHandy();
```
ry: `makeHandy` should not modify `.bash_profile` it should only set the exe to be executable.

install.ts
chiefbiiko: thanks for the hint! :100:
thought I would also check the process status success prop. makes sense?

install.ts
```diff
@@ -0,0 +1,234 @@
...
+  child.close();
+  if (!isWindows) {
+    if (platform.os === "linux") {
+      await rename(DENO_BIN, OLD_DENO_BIN);
+    }
+    await copyFile(archive.replace(/\.gz$/, ""), DENO_BIN);
+  }
+}
+
+async function makeHandy(): Promise<void> {
```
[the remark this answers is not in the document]
chiefbiiko: thanks for your feedback!
will toss all bash_profile related stuff.
should the script do the same on windows or have powershell update the PATH still?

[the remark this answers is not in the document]
MarkTiedemann: The new PowerShell script updates the Path on Windows, but does nothing on Unix.

install.ts
```diff
@@ -0,0 +1,218 @@
...
+  exit,
+  platform,
+  writeFile,
+  removeAll,
+  rename,
+  run,
+  makeTempDir
+} from "deno";
+
+import { mkdirp } from "https://deno.land/x/std/fs/mkdirp.ts";
```
MarkTiedemann: I read `deno.mkdir` is already recursive. Maybe this can be removed?

See: [https://github.com/denoland/deno_install/pull/21/files](https://github.com/denoland/deno_install/pull/21/files(

README.md
```diff
@@ -20,6 +20,14 @@ iex (iwr https://deno.land/x/install/install.ps1)
 
 _Note: Depending on your security settings, you may have to run `Set-ExecutionPolicy RemoteSigned -Scope CurrentUser` first to allow downloaded scripts to be executed._
 
+**Update with deno:**
+
+```
+deno --allow-write --allow-net --allow-env --allow-run https://deno.land/x/install/install.ts
```
satyarohith: We can use `--allow-all` with new version `0.2.7`.

install.ts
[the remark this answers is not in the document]
chiefbiiko: @MarkTiedemann Thanks doing the same here now.

install.ts
chiefbiiko: @MarkTiedemann Wow, that's cool!
That makes `mkdirp` obsolete indeed.

PR #23, opened 2019-01-10
Add install.sh

MarkTiedemann:
Started working on a `.sh` installer script.

I don't have much `bash` experience so any feedback is much appreciated.

____

- [x] Add tests
- [x] Update docs


install.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ $# -eq 0 ]; then
+  # TODO: We could check the GitHub API for the latest release here.
+  # However, we need to parse the JSON response body. Perhaps
+  # this could be done with an existing installation of deno, jq,
+  # or node. And otherwise we ned to fall back to asking the user:
+  read -p "Which version do you want to install? " version
+else
+  version=$1
+fi
+
```
chiefbiiko: @MarkTiedemann check how I determine the release url in [function release_url](https://github.com/chiefbiiko/deno-bash-installer/blob/63aada0db1792b20417bd4a4dcb5865a90cf8929/install.sh#L31) and [function main (l57-l59)](https://github.com/chiefbiiko/deno-bash-installer/blob/63aada0db1792b20417bd4a4dcb5865a90cf8929/install.sh#L57)
if u pass a tag that version is installed else the latest
this line `match=$(curl -sL $url | grep href | grep $filename | grep -o '/deno[^"]*')` has been quite reliable so far
the `grep`s work on mac and linux
just use a remix of `release_url` and that regex in `main`

MarkTiedemann: @chiefbiiko You're grepping the HTML of the release page? This could break anytime GitHub is updating their markup, so maybe it "has been quite reliable so far", but I wouldn't want to use that. Instead, I'd rather grep or json-parse the GitHub API response because, at the very least, the API has some stability guarantee.

For example: `curl -s https://api.github.com/repos/denoland/deno/releases/latest | grep tag_name | grep -o "v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}"`

Then again, I'm not doing that on purpose to avoid complexity (see: [https://github.com/denoland/deno_std/issues/94#issuecomment-451981613](https://github.com/denoland/deno_std/issues/94#issuecomment-451981613)).

I'd be interested to know what @ry thinks about this. In general, we should decide whether we want to hard-code the version number or parse it from the GitHub API, and then implement that in all the different scripts.

ry: I'm okay with hardcoding it if we move this to deno_std. 

We are being very careful in our release process to make it as automated as possible. Currently the only manual steps for release are:
1. make a commit in deno bumping the version
2. push a tag
3. [update the CI scripts in deno_std](https://github.com/denoland/deno_std/commit/72b2e20f9da1d431ba9cac3247ee38ee82d5d8f4) 

Step 3 is annoying, but I want deno_std to be explicitly testing against a specific version. I don't want to also add a step 4 "Bump deno_install"

Another reason in favor of hard coding is that sometimes the Windows binaries are cut faster than the Linux or Mac binaries (because appveyor is faster) ... so for a short time there are no Linux and Mac releases of the "latest version". If we manually bumped it, it wouldn't be a problem.

Ideally I'd be able to script the action of bumping the deno_std version, so that the release process is minimal.

install.sh
```diff
@@ -0,0 +1,60 @@
...
+green="\033[32m"
+yellow="\033[33m"
+reset="\033[0m"
+
+if [ $# -eq 0 ]; then
+  # TODO: We could check the GitHub API for the latest release here.
+  # However, we need to parse the JSON response body. Perhaps
+  # this could be done with an existing installation of deno, jq,
+  # or node. And otherwise we ned to fall back to asking the user:
+  read -p "Which version do you want to install? " version
```
ry: This isn't helpful... The point of these scripts is to be a single command install.

install.sh
MarkTiedemann: > I don't want to also add a step 4 "Bump deno_install"

Agreed.

> Another reason in favor of hard coding is that sometimes the Windows binaries are cut faster than the Linux or Mac binaries (because appveyor is faster) ... so for a short time there are no Linux and Mac releases of the "latest version".

That's interesting! I didn't know that.

Looking at the `created_at` fields of the assets in the API response, "a short time" can actually mean up to 20 minutes in this case.

Hard-coding is fine with me. Alternatively, we could also get the last 2 releases from the API and use the first available `browser_download_url` for the OS the script is being run on.

```
$ curl -s https://api.github.com/repos/denoland/deno/releases?per_page=2 | grep browser_download_url
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.6/deno_linux_x64.gz"
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.6/deno_osx_x64.gz"
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.6/deno_win_x64.zip"
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.5/deno_linux_x64.gz"
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.5/deno_osx_x64.gz"
"browser_download_url": "https://github.com/denoland/deno/releases/download/v0.2.5/deno_win_x64.zip"
```

I guess it should be possible to implement this for the deno_std CI, too, right? If so, step 3 would not be needed anymore. :)

install.sh
MarkTiedemann: Yeah, that's kinda just a placeholder, until we decide on how to proceed with the "how to get the latest version?" problem.

install.sh
MarkTiedemann: Hard-coded the $version for now. I guess if we want to do it dynamically or move this code to deno_std, we can do so later.

MarkTiedemann:
> Please analyse it with shellcheck and address those warnings. It complains mostly about not quoting command arguments tho. 👍

Thanks! I didn't know about `shellcheck`.

> Also it would be nice if you can make it POSIX compatible meaning you use `sh` rather than `bash`isms so BSD users can leverage it too.

I'm new to bash-scripting. What exactly do I need to change? Can you do a more detailed review, please?

MarkTiedemann:
Should I use `chmod u+x` instead of `+x`?

install.sh
```diff
@@ -0,0 +1,34 @@
...
+
+curl -fL# -o "$deno_dir/deno.gz" "$deno_uri"
+gunzip -df "$deno_dir/deno.gz"
+chmod +x "$deno_dir/deno"
+
+echo "Deno was installed successfully."
+case $PATH in
+  *"$deno_dir"*) echo "Run 'deno --help' to get started." ;;
+  *) echo "Run '~/.deno/bin/deno --help' to get started." ;;
+esac
```
vladimyr: For extra safety wrap those paths with `:`:
```sh
case ":$PATH:" in
  *:$deno_dir:*) echo "Run 'deno --help' to get started." ;;
  *) echo "Run '~/.deno/bin/deno --help' to get started." ;;
esac
```

install.sh
```diff
@@ -0,0 +1,34 @@
...
+deno_dir="$HOME/.deno/bin"
+
+if [ ! -d "$deno_dir" ]; then
+  mkdir -p "$deno_dir"
+fi
+
+case $(uname -s) in
+  Darwin) os="osx" ;;
+  *) os="linux" ;;
+esac
```
vladimyr: You might lowercase it if you want:
```sh
case "$(uname -s | tr '[:upper:]' '[:lower:]')" in
  darwin) os="osx";;
  *) os="linux";;
esac
```

install.sh
MarkTiedemann: Wrapping in `:` won't help if it's the last or first path.

Actually I'd need to check `$HOME/.deno/bin`, `$HOME/.deno/bin/`, `$HOME/.deno/bin:*`, `$HOME/.deno/bin/:*`, `*:$HOME/.deno/bin:*`, `*:$HOME/.deno/bin/:*`, `*:$HOME/.deno/bin`, `*:$HOME/.deno/bin/`, and all of those variants with `$HOME` being replaced by `~`.

That would be perfectly correct, but hard to maintain. I'd prefer to just check `*$HOME/.deno/bin*` and `*~/.deno/bin*` for the sake of maintainability.

vladimyr: > Wrapping in `:` won't help if it's the last or first path.

I think you missed that `":$PATH:"` is wrapped too, not just `*:$deno_dir:*`?

MarkTiedemann:
> Keep in mind that +x means a+x

I know that, but I don't know what we want, as in "What's the best practice regarding permissions here?".

install.sh
MarkTiedemann: In the `pwsh` script, I split the PATH by `:` and then check the splitted paths for `$HOME/.deno/bin`, `$HOME/.deno/bin/`, `~/.deno/bin` and `~/.deno/bin/`. I just don't know how to do this nicely in `sh`.

vladimyr: Checking for `~/.deno/bin` seems awkward 🙃
Only way for unexpanded `~` to get inside path is doing something like this:
```sh
PATH="~/.deno/bin":$PATH
```
by specifically instructing shell not to expand path by quoting it. I firmly believe this may end up in your path only by mistake and we shouldn't account for _wrongly_ declared `PATH`s?

vladimyr: > In the `pwsh` script, I split the PATH by `:` and then check the splitted paths for `$HOME/.deno/bin`, `$HOME/.deno/bin/`, `~/.deno/bin` and `~/.deno/bin/`. I just don't know how to do this nicely in `sh`.

You are looking it from different perspective. I'm not talking about splitting and checking if resulting array contains matches. What I did in my snippet is:
1. wrapping original `PATH` value with surrounding `:`
2. finding `:<target>:` substring in result of 1.

Reason why would you like to wrap is because without it you are matching against some portion of a path or a full one:
```sh
*$deno_dir*) ;;
# would match: `/usr/bin:$deno_dir:/usr/local/bin`
# but also: `/usr/bin:$deno_dir/some/trailing/stuff:/usr/local/bin`
# and this ofc: `/usr/bin:/parent/dirs/$deno_dir:/usr/local/bin`
```
while:
```sh
*:$deno_dir:*) ;;
# would match: `/usr/bin:$deno_dir:/usr/local/bin`
# would NOT match: `/usr/bin:$deno_dir/some/trailing/stuff:/usr/local/bin`
```

MarkTiedemann: > I firmly believe this may end up in your path only by mistake

I see. I'll fix that.

> wrapping original PATH value with surrounding :

I missed that. Thanks!

MarkTiedemann: @vladimyr Am I doing the wrong check maybe? Would it perhaps be better to check whether deno is available on the PATH by running `which deno`?

MarkTiedemann: I used `type` instead of `which` now. What do you think, @vladimyr?

vladimyr: From my experience, reading other projects installation scripts, people typically use `command -v <command>` but I don't think using `type` instead is bad either.

MarkTiedemann: Thank you, @vladimyr! Fixed it.

MarkTiedemann:
I added a test file and updated the docs.

Not quite sure how to structure the `.travis.yml` file so that it runs the `python` 2.7 and python 3.3 tests on Linux as well as the `sh` test on Linux and MacOS.

Can anyone help with that?

.travis.yml
```diff
@@ -1,16 +1,55 @@
-language: python
+matrix:
+  include:
+    - language: python
+      os:
+        - linux
+        # Travis does not support python on osx yet.
+        # see: https://github.com/travis-ci/travis-ci/issues/2312
+        # - osx
```
ry: Really? We should look into using azure-pipelines... [It's working well in deno_std](https://github.com/denoland/deno_std/blob/master/azure-pipelines.yml)

.travis.yml
```diff
@@ -1,16 +1,55 @@
...
+        - shellcheck --version
+        - curl -sSL "$SHFMT_DOWNLOAD_URL" -o $HOME/bin/shfmt;
+          chmod +x $HOME/bin/shfmt;
+          shfmt --version
+      script:
+        - echo "Linting installer script using \`shellcheck\`:";
+          shellcheck -s sh *.sh;
+          shellcheck -s dash *.sh;
+          shellcheck -s ksh *.sh;
+          shellcheck -s bash *.sh
```
ry: ok nice.

.travis.yml
MarkTiedemann: Well, but this shouldn't block this PR, right?

I think we should move to azure-pipelines if this repo is merged into deno_std.

MarkTiedemann:
> I will merge this repo into deno_std after we land install.ts PR.

Does this mean the hard-coding decision is final?

If so, have you considered the dynamic approach I mentioned? Getting the last two releases from the GitHub API and then using the first available asset for the download? Might not be perfect for `sh` (due to lack of JSON parsing) but wouldn't be fragile with well-implemented `grep`ping either. In any case, for `pwsh`, `deno` and `python`, this would be easy to do, and would spare you from having to update so many hard-coded version numbers. Also, it's pretty fail-safe in case you forget to update the hard-coded version, or publish the hard-coded version before all assets are available. In fact, if we only used that approach, you wouldn't have to worry about updating the hard-coded versions nor missing assets when users try to download the latest release.

ry:
If you can figure out how to do it with just unix tools like grep, cut, awk even, I'd be happy to land it.

The idea with installer scripts is to be as portable as possible - and now that we've got a pure shell script and a powershell script - I will start deprecating the python one.

vladimyr:
@MarkTiedemann gh's _hidden_ gem: https://github.com/denoland/deno/releases.atom
Asset names are fixed and asset download url format is known:
```
https://github.com/denoland/deno/releases/download/<version>/<asset>
```
all you need to do is extract versions from xml (you'll get more than enough obviously). It is _guaranteed_ to be pretty printed so that makes your regex construction easier. Also you don't need to resort to regex right away. There is awesome `xmllint` tool that comes preinstalled on macos and bunch of linux distros so check for that first 😉

install_test.sh
```diff
@@ -0,0 +1,40 @@
...
+  branch=$TRAVIS_PULL_REQUEST_BRANCH
+fi
+e2e_url=https://raw.githubusercontent.com/$repo/$branch/install.sh
+
+rm -rf ~/.deno
+curl -sSL "$e2e_url" | $SHELL
+deno -v
+
+rm -rf ~/.deno
+curl -sSL "$e2e_url" | $SHELL -s v0.2.0
```
vladimyr: @MarkTiedemann 
I went here to comment on `./install.sh` invocations I completely missed earlier 🤦‍♂️ 
They should be written as `$SHELL ./install.sh` but then I found line above and that make my day: 🏆 👍x 💯

MarkTiedemann:
> Asset names are fixed and asset download url format is known:

The problem is that, once a release is published, it is not guaranteed that all assets have been added by the CI systems. For example, it can take 20 minutes for all binaries to be added to the release.

Luckily, the GitHub API release object has a list of the assets that have been created already. So what I'm proposing is to get the last two releases. And if the newest release doesn't have an asset yet, then we just download the asset from the second-newest release.

vladimyr:
I'm aware of asset releasing delay. Saw @ry's mentioning it somewhere.

In my `get-deno` node package I did use gh api so I know what you are talking about :wink:

With that being said you can use http as your api (as in my fictional example): `https://github.com/denoland/deno/releases/download/v0.2.6/deno_freebsd_x64.gz`. That will give you 404 as expected and you can fallback to previos version then. Relevant SO example: https://stackoverflow.com/q/24966975

MarkTiedemann:
I know. But then you need to know the previous version. If the version is "1.0.0", the previous version could be anything from " 0.0.0" to "0.9999999.9999999". Why do multiple HTTP calls when we can get all the info that we need in 1 call?

vladimyr:
Because atom feed will get you all versions so you'll knew exactly what's previous one. And that means 2 calls on good day, 3 on a bad one (because of fallback). In consuming gh api scenario it still means 2 calls on a good day because you need to fetch releases first?

MarkTiedemann:
> In consuming gh api scenario it still means 2 calls on a good day because you need to fetch releases first?

No, luckily just 1 call. :) The call I mentioned earlier (`https://api.github.com/repos/denoland/deno/releases?per_page=2`) returns the last two releases. Each of those releases has an assets property which is an array containing all the assets of that release.

How widely available is `xmllint`? I just heard about it for the first time now.

vladimyr:
Depends what do you count, I counted download call to :wink:
Well macos, some linux distros but you'll do self a favor avoiding it and using plain grep (tried after posting comment).

MarkTiedemann:
Plain grep it is. :)

Can I tag you for a review once I'm done?

MarkTiedemann:
```
$ os=win
$ zip=zip
$ curl -sS https://api.github.com/repos/denoland/deno/releases?per_page=2 | grep -o "https://github.com/denoland/deno/releases/download/v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}/deno_${os}_x64\.${zip}"
https://github.com/denoland/deno/releases/download/v0.2.7/deno_win_x64.zip
https://github.com/denoland/deno/releases/download/v0.2.6/deno_win_x64.zip
```
I think this might be stable enough already. What do you think, @vladimyr?

vladimyr:
@MarkTiedemann Sorry for being unresponsive. I end up being pretty busy with offline stuff last few days. FWIW this does look good 👍

PR #25, opened 2019-01-11
Rewrite PowerShell script to be cross-platform

MarkTiedemann:
- [x] Make the script work on `linux` and `osx`
- [x] Harmonize with `sh` installer
- [x] Update tests

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -15,23 +15,23 @@ if ($PSVersionTable.PSVersion.Major -lt 6) {
 }
 
 $DenoDir = if ($IsWindows)
-  { "${Home}\.deno\bin" } else
-  { "${Home}/.deno/bin" }
+  { "$Home\.deno\bin" } else
+  { "$Home/.deno/bin" }
 
 $Zip = if ($IsWindows)
   { 'zip' } else
   { 'gz' }
 
 $DenoZip = if ($IsWindows)
-  { "${DenoDir}\deno.${Zip}" } else
-  { "${DenoDir}/deno.${Zip}" }
+  { "$DenoDir\deno.$Zip" } else
+  { "$DenoDir/deno.$Zip" }
 
 $OS = if ($IsWindows)
   { 'win' } else { if ($IsMacOS)
   { 'osx' } else
   { 'linux' } }
 
-$DenoUri = "https://github.com/denoland/deno/releases/download/${Version}/deno_${OS}_x64.${Zip}"
+$DenoUri = "https://github.com/denoland/deno/releases/download/$Version/deno_${OS}_x64.$Zip"
 
 if (!(Test-Path $DenoDir)) {
   New-Item $DenoDir -ItemType Directory | Out-Null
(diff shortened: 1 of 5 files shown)


MarkTiedemann:
Oh, and I removed the call to the GitHub API to get the latest version. @ry Are you fine with hard-coding this?

MarkTiedemann:
On another note, I kept the code adding deno to the Path on Windows, since it's easily doable with the APIs there.

But doing it for MacOS and Linux seemed more of a hassle than just printing `Run '~/.deno/bin/deno --help' to get started.`. Alternatively, I'd rather symlink to `/usr/bin` or something than add to the PATH on MacOS and Linux - I'd prefer not to mess with anyone's `.bash_profile`. So yeah, I think not doing it and printing the command with the full path is the best and simplest solution for now.

MarkTiedemann:
On second thought, I'm not sure whether implementing gzip-unzipping via standard library instead of shelling out to `gunzip` is a good idea... That stream read-write code has a much higher complexity than a simple `gunzip -d $DenoZip`.

install.ps1
```diff
@@ -1,88 +1,66 @@
...
-  Write-Host $Text -NoNewline
-}
-function Write-Emphasized ([string] $Text) {
-  Write-Host $Text -NoNewLine -ForegroundColor "Yellow"
-}
-function Write-Done {
-  Write-Host " done" -NoNewline -ForegroundColor "Green";
-  Write-Host "."
+if ($PSVersionTable.PSVersion.Major -lt 6) {
+  $IsWindows = $true
```
piscisaureus: Are $IsWindows and $IsMacOS powershell builtins? 
Otherwise I don't understand this logic -- I'm pretty sure it's possible to have PS 6 on windows.

install.ps1
```diff
@@ -1,88 +1,66 @@
...
-
-# Download Deno release.
-$zip_file = "${deno_dir}\deno_win_x64.zip"
-$download_uri = "https://github.com/denoland/deno/releases/download/" +
-                "${version}/deno_win_x64.zip"
-Write-Part "Downloading "; Write-Emphasized $download_uri; Write-Part "..."
-Invoke-WebRequest -Uri $download_uri -OutFile $zip_file
-Write-Done
+$DenoZip = if ($IsWindows)
+  { "$DenoDir\deno.$Zip" } else
```
piscisaureus: I think a forward slash works fine on Windows too.

install.ps1
```diff
@@ -1,88 +1,66 @@
...
+  $IsMacOS = $false
 }
 
-if (-not $version) {
-  # Determine latest Deno release via GitHub API.
-  $latest_release_uri = "https://api.github.com/repos/denoland/deno/releases/latest"
-  Write-Part "Downloading "; Write-Emphasized $latest_release_uri; Write-Part "..."
-  $latest_release_json = Invoke-WebRequest -Uri $latest_release_uri
-  Write-Done
+$DenoDir = if ($IsWindows)
```
piscisaureus: I'd prefer that you use "normal" formatting for if blocks.

install.ps1
MarkTiedemann: Yes, `$IsWindows`, `$IsMacOS` and `$IsLinux` are built-ins in PowerShell Core (starting from PowerShell v6). So for <v6, I'm adding the built-ins.

MarkTiedemann:
> | Are you fine with hard-coding this?
> Why did you remove that? We want to update it every time a new version of deno is released.

- [https://github.com/denoland/deno_install/pull/23#discussion_r247363012](https://github.com/denoland/deno_install/pull/23#discussion_r247363012)
- [https://github.com/denoland/deno_std/issues/94#issuecomment-451981613](https://github.com/denoland/deno_std/issues/94#issuecomment-451981613)

I'm fine with both hardcoding the version or fixing the API calls to account for missing assets.

install.ps1
MarkTiedemann: I'm not sure. Have you tested this with both PowerShell and PowerShell Core on Windows? If so, I'd gladly change it. :)

Unfortunately, I only will have access to my Windows machine on Saturday again.

install.ps1
MarkTiedemann: Hmm, I tried to make it look like an [iif](https://en.wikipedia.org/wiki/IIf).

How about?

```powershell
if ($IsWindows) {
  $DenoDir = "$Home\.deno\bin"
} else {
  $DenoDir = "$Home/.deno/bin"
}
```

install.ps1
MarkTiedemann: Maybe we should keep the backslashes on Windows since the path of the executable will be printed?

MarkTiedemann:
Apparently, this beautiful piece of code is not working in `powershell`, but works just fine in `pwsh`.

```powershell
(Invoke-WebRequest 'https://github.com/denoland/deno/releases').Links |
    Where-Object { $_.href -like "/denoland/deno/releases/download/*/deno_${OS}_x64.$Zip" } |
    ForEach-Object { $_.href } |
    Select-Object -First 1
```

My apologies for getting the CI stuck a couple of times. Unfortunately, I don't have access to a Windows machine at the moment. So I kinda have to debug in CI...

Will try to fix this tomorrow some time. Hopefully, this PR can be merged soon, in the coming days.

MarkTiedemann:
@piscisaureus Can you review again, please?

MarkTiedemann:
@ry Once this PR is merged, we could use the `sh` and `powershell` installers in the `deno_std` CI so you no longer need to update the `DENO_VERSION` manually after each release.

MarkTiedemann:
> I didn’t realize one could run powershell outside of windows.

Haha, yeah, only since PowerShell v6 a.k.a. PowerShell Core. :) It's based on the new cross-platform, open-source .NET Core.

Summpot:
Reason? Because powershell supports cross-platform, and this script has no obvious label that it cannot cross-platform (I planned to use this script on linux at the beginning).
As for the rate limit of the Github API,
I think that normal users are unlikely to trigger this limit.

PR #27, opened 2019-01-15
Add instructions and bump to v0.2.7

ry:


install.sh
```diff
@@ -5,12 +5,13 @@
...
 
 if [ $# -eq 0 ]; then
-  version="v0.2.6"
+  version="v0.2.7"
 else
   version=$1
 fi
 
 deno_dir="$HOME/.deno/bin"
+exe="$deno_dir/deno"
```
MarkTiedemann: Maybe `deno_exe`? Or rename `deno_dir` `dir` for variable name consistency?

install.sh
```diff
@@ -5,12 +5,13 @@
 set -e
 
 if [ $# -eq 0 ]; then
-  version="v0.2.6"
+  version="v0.2.7"
```
piscisaureus: I Always assumed the python installer did auto detection.

install.sh
```diff
@@ -23,13 +24,15 @@ esac
...
+chmod +x "$exe"
 
-echo "Deno was installed successfully."
+echo "Deno was installed successfully to $exe"
 if command -v deno >/dev/null; then
   echo "Run 'deno --help' to get started."
 else
-  echo "Run '~/.deno/bin/deno --help' to get started."
+  echo "You must manually add it to your path:"
+  echo "  export PATH=\"$deno_dir:\$PATH\""
```
MarkTiedemann: Obviously, this won't persist the PATH variable. But should that be mentioned, too?

I specifically didn't mention how to add deno to the PATH nor how to persist it since I think the target audience is familiar with this.

install.sh
```diff
@@ -23,13 +24,15 @@ esac
...
+gunzip -df "$exe.gz"
+chmod +x "$exe"
 
-echo "Deno was installed successfully."
+echo "Deno was installed successfully to $exe"
 if command -v deno >/dev/null; then
   echo "Run 'deno --help' to get started."
 else
-  echo "Run '~/.deno/bin/deno --help' to get started."
+  echo "You must manually add it to your path:"
```
MarkTiedemann: Some people may choose to symlink to another directory already in the PATH. I think "you must add it to your PATH" sounds kinda weird. The "must" specifically makes you think "so if I don't add deno to the PATH, will it not work?".

install.sh
ry: The python one does. This one has it hardcoded for now.

install.sh
ry: Fixed

install.sh
ry: Noted.

install.sh
```diff
@@ -23,13 +24,15 @@ esac
...
 
-curl -fL# -o "$deno_dir/deno.gz" "$deno_uri"
-gunzip -df "$deno_dir/deno.gz"
-chmod +x "$deno_dir/deno"
+curl -fL# -o "$exe.gz" "$deno_uri"
+gunzip -df "$exe.gz"
+chmod +x "$exe"
 
-echo "Deno was installed successfully."
+echo "Deno was installed successfully to $exe"
```
MarkTiedemann: The `.` at the end of the sentence was removed. The next sentence ends with a `.`, too. So for consistency, I would re-add the `.`.

install_test.sh
```diff
@@ -1,6 +1,18 @@
...
+shellcheck -s sh ./*.sh;
+shellcheck -s dash ./*.sh;
+shellcheck -s ksh ./*.sh;
+shellcheck -s bash ./*.sh
+shfmt -d -p -i 2 -ci .
+sh ./install_test.sh;
+dash ./install_test.sh;
+ksh ./install_test.sh;
+bash ./install_test.sh;
+zsh ./install_test.sh
```
MarkTiedemann: Nit, the trailing `;`s can be removed.

.travis.yml
```diff
@@ -14,7 +14,7 @@ matrix:
         - ./install.py
         - ./install_test.py
 
-    - language: sh
+    - language: bash
```
MarkTiedemann: This could also be `shell` or `minimal`.

See: [https://docs.travis-ci.com/user/languages/minimal-and-generic/#aliases](https://docs.travis-ci.com/user/languages/minimal-and-generic/#aliases)

Personally, I liked `sh`, but I'm fine with `bash`, too. ;)

install.sh
ry: yes, makes it more copy & paste-able

.travis.yml
vladimyr: I'm with you in this one. Liked `sh` more. 🙄

vladimyr:
@ry Whats rationale behind removing actual `install.sh` invocations in various `sh`-like shells and e2e test?

PR #29, opened 2019-01-18
Remove hardcoded version

MarkTiedemann:


install.sh
```diff
@@ -4,10 +4,17 @@
...
 
+case $(uname -s) in
+Darwin) os="osx" ;;
+*) os="linux" ;;
+esac
+
 if [ $# -eq 0 ]; then
-	version="v0.2.7"
+	deno_uri=$(curl -sS https://api.github.com/repos/denoland/deno/releases?per_page=2 |
+		grep -o "https://github.com/denoland/deno/releases/download/v[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}/deno_${os}_x64\.gz" |
```
ry: Probably `"https://github.com/denoland/deno/releases/download/\.*\.gz"` works

MarkTiedemann: Almost. Without the backslash it seems to work fine.

```diff
+"https://github.com/denoland/deno/releases/download/.*\.gz"
-"https://github.com/denoland/deno/releases/download/\.*\.gz"
```

MarkTiedemann: Actually, no, that doesn't work - It will take the `linux` version on `osx` then.

Will use: `"https://github.com/denoland/deno/releases/download/.*/deno_${os}_x64\.gz"`

MarkTiedemann:
The anonymous requests to the GitHub API are hitting the ratelimit; probably because the CI jobs are running on a shared server. Since the `.sh` installer script might be used by other people in CI scripts, too, I think this is an important use case, and I kept a hardcoded fallback to the last know version.

What do you think? @ry

MarkTiedemann:
On a related note, could we POST the deno releases as drafts first and then PATCH them when all assets have been added? (See `draft` poperty: [https://developer.github.com/v3/repos/releases/#edit-a-release](https://developer.github.com/v3/repos/releases/#edit-a-release))

(That's of course not solving the ratelimiting problem, but at least we wouldn't need to get the last two versions...)

MarkTiedemann:
We could also just use the actual HTML release page.

`$ echo "https://github.com$(curl -sSf https://github.com/denoland/deno/releases | grep -o "/denoland/deno/releases/download/.*/deno_osx_x64\.gz" | head -n 1)"`
`https://github.com/denoland/deno/releases/download/v0.2.7/deno_osx_x64.gz`

This seems like the best solution to me since we don't have any trouble with the ratelimiting then.

ry:
> This seems like the best solution to me since we don't have any trouble with the ratelimiting then.

Oh yeah, I forgot but I hit this rate limit problem early on with the python installer. Hitting the html page is better.

install.sh
```diff
@@ -4,10 +4,19 @@
...
+*) os="linux" ;;
+esac
+
 if [ $# -eq 0 ]; then
-	version="v0.2.7"
+	deno_uri="https://github.com$(
+		curl -sSf https://github.com/denoland/deno/releases |
+			grep -o "/denoland/deno/releases/download/.*/deno_${os}_x64\.gz" |
+			head -n 1
+	)"
```
ry: Cool - glad you got this to work.

MarkTiedemann:
I'll fix the error handling (unfortunately, `set -o pipefail` is not working in `dash` so I have to get rid of the pipes).

ry:
@MarkTiedemann Just make it work in 'sh'

MarkTiedemann:
> @MarkTiedemann Just make it work in 'sh' - it's the lowest common denominator

`/bin/sh` on Ubuntu (and in CI) is symlinking `/bin/dash` by default. I'd like to support Ubuntu.

MarkTiedemann:
Pretty happy with the PR now. :)

install.sh
```diff
@@ -4,10 +4,19 @@
 
 set -e
 
+case $(uname -s) in
+Darwin) os="osx" ;;
+*) os="linux" ;;
```
vladimyr: Hmm, it would be safer to check for `Linux` here 🤔

install.sh
vladimyr: > Probably `"https://github.com/denoland/deno/releases/download/\.*\.gz"` works

But why? Why lose all meaning? 😞

Issue #30, opened 2019-01-18
Improve CI


MarkTiedemann:
- [x] Re-add testing the `sh` installer on `osx`
- [ ] Re-add testing the `sh` installer with different shells (`ksh`, `zsh`, `dash`, `bash`, and `sh`)
- [x] Once the `pwsh` installer is cross-platform, add `win`, `osx` and `linux` tests
- [ ] Once the `sh` installer is cross-platform, add `win` test

MarkTiedemann:
`zsh` and `ksh` tests were not deemed necessary.

Closed by [#31](https://github.com/denoland/deno_install/pull/31) and [#25](https://github.com/denoland/deno_install/pull/25).

Bash for Windows support is covered by [#28](https://github.com/denoland/deno_install/issues/28).

PR #31, opened 2019-01-18
Re-add CI for osx

MarkTiedemann:
- Re-add testing the `sh` installer on `osx`
- ~~Re-add testing the `sh` installer with different shells (`ksh`, `zsh`, `dash`, `bash`, and `sh`)~~

See: [#30](https://github.com/denoland/deno_install/issues/30).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ matrix:
       python:
         - "2.7"
         - "3.3"
-      # Test the install script.
       script:
         - ./install.py
         - ./install_test.py
@@ -17,21 +16,15 @@ matrix:
     - language: bash
       os:
         - linux
+        - osx
       addons:
         apt:
           packages:
             - ksh
             - zsh
-      env:
-        - PATH=$HOME/bin/:$PATH
-        - SHFMT_VERSION=v2.6.2
-        - SHFMT_DOWNLOAD_URL="https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_linux_amd64"
-      before_script:
-        - $SHELL --version 2>/dev/null || dpkg -s "$SHELL" 2>/dev/null || command -v "$SHELL"
-        - curl --version
-        - shellcheck --version
-        - curl -sSL "$SHFMT_DOWNLOAD_URL" -o $HOME/bin/shfmt;
-        - chmod +x $HOME/bin/shfmt;
-        - shfmt --version
+        homebrew:
+          packages:
+            - ksh
+            - zsh
       script:
         - ./install_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,18 @@ matrix:
     - language: bash
       os:
         - linux
-        - osx
       addons:
         apt:
           packages:
             - ksh
             - zsh
+      script:
+        - ./install_test.sh
+
+    - language: bash
+      os:
+        - osx
+      addons:
         homebrew:
           packages:
             - ksh
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,12 +40,6 @@ matrix:
             - ksh
             - zsh
             - shellcheck
-      env:
-        - PATH=/usr/local/bin:$PATH
-        - SHFMT_OS=darwin
-        - SHFMT_VERSION=v2.6.2
-      before_script:
-        - curl -sSL -o "/usr/local/bin/shfmt" "https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_${SHFMT_OS}_amd64"
-        - chmod +x "/usr/local/bin/shfmt"
+            - shfmt
       script:
         - ./install_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
         - SHFMT_OS=linux
         - SHFMT_VERSION=v2.6.2
       before_script:
-        - curl -sSL -o "$HOME/bin/shfmt" https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_${SHFMT_OS}_amd64";
+        - curl -sSL -o "$HOME/bin/shfmt" "https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_${SHFMT_OS}_amd64";
         - chmod +x "$HOME/bin/shfmt"
       script:
         - ./install_test.sh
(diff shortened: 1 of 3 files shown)


install_test.sh
```diff
@@ -1,17 +1,52 @@
...
+*) shells=(sh bash dash ksh zsh) ;;
+esac
 
-# Test we can get the latest version.
-rm -rf ~/.deno
-sh ./install.sh
-~/.deno/bin/deno -v
+for shell in $"${shells[@]}"; do
+	test_specific_version $shell
+	test_latest_version $shell
```
ry: This isn't necessary - we only need to test under sh.

MarkTiedemann: But `sh` on `osx` is just `bash`, right?

And `dash`, for example, doesn't support `set -o pipefail`.

Is this really not necessary?

MarkTiedemann: Perhaps it's not necessary, but it also doesn't hurt to test a bit more, right? :)

MarkTiedemann: Right now, again, CI is failing on `osx`, but working on `linux`. Yesterday, `linux` was failing for me, but `osx` was working.

I'd rather test more options than less. :)

install_test.sh
```diff
@@ -1,17 +1,44 @@
...
+# Check formatting.
+./shfmt -d .
 
-# Test we can get the latest version.
-rm -rf ~/.deno
-sh ./install.sh
-~/.deno/bin/deno -v
+# Lint code.
+shellcheck -s bash ./*.sh
+shellcheck -s sh ./install.sh
```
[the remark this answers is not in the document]
MarkTiedemann: Okay.

MarkTiedemann:
Kinda happy with the PR now.

@ry Hopefully, I'll be able to work on Bash for Windows support tomorrow - that's probably gonna be a CI update again.

install_test.sh
```diff
@@ -1,17 +1,42 @@
...
-shellcheck -s sh ./*.sh
-shfmt -d .
+# Install shfmt.
+shfmt_version=v2.6.2
+case $(uname -s) in
+Darwin) shfmt_os="darwin" ;;
+*) shfmt_os="linux" ;;
+esac
+shfmt_url="https://github.com/mvdan/sh/releases/download/${shfmt_version}/shfmt_${shfmt_version}_${shfmt_os}_amd64"
+curl -sSL -o ./shfmt "$shfmt_url"
```
ry: I don't want to download shfmt when I run it locally. Can you leave that in the travis script?

install_test.sh
```diff
@@ -1,17 +1,42 @@
-#!/bin/sh
```
ry: I think this file should be reverted to its original state

install.sh
```diff
@@ -11,7 +11,7 @@ esac
 
 if [ $# -eq 0 ]; then
 	deno_asset_path=$(curl -sSf https://github.com/denoland/deno/releases |
-		grep -o "/denoland/deno/releases/download/.*/deno_${os}_x64\.gz" |
+		grep -o "/denoland/deno/releases/download/.*/deno_${os}_x64\\.gz" |
```
ry: Intentional?

.travis.yml
```diff
@@ -22,16 +21,17 @@ matrix:
...
+        - ./install_test.sh
+
+    - language: bash
+      os:
+        - osx
+      addons:
+        homebrew:
+          packages:
+            - ksh
+            - zsh
```
ry: ksh and zsh are not needed.

install_test.sh
MarkTiedemann: I didn't run this test locally. For a local test I always just ran `./install.sh && ~/.deno/bin/deno -v`.

But sure, I can move it back if you want.

install.sh
MarkTiedemann: Yes, a newer version of `shellcheck` reported this, and it's correct.

.travis.yml
MarkTiedemann: Why are they not needed?

Are they not supported?

Do they not need to be tested for some technical reason?

I don't understand.

MarkTiedemann: Okay, but running `sh` in CI on `linux` is not running `sh` but `dash`. And running `sh` in CI on `osx` is `bash`.

Is every script that is gonna run in `dash` gonna run fine in `zsh`, too? Is every script that is gonna run in `bash` gonna run fine in `zsh`, too?

Is there something I'm missing here? I'm kinda new to shell-scripting.

install_test.sh
MarkTiedemann: In the .yaml file, there'd be duplication since there are seperate sections for osx and linux; in the script, it's just a case statement.

Would it be okay if I put the shfmt download in a "if CI" block?

ry: Ok, sure - but use this:

```sh
if ! command -v shfmt; then
  echo "shfmt not installed"
fi
```

MarkTiedemann: Actually, I removed it from the script now and put it into the .travis.yml file since the installation locations on linux and osx are different.

install_test.sh
MarkTiedemann: All of it? Or should I just remove the different shells? What about the part where I added the osx CI?

.travis.yml
ry: "sh" is the lowest common denominator - if it works under that it'll work.

MarkTiedemann: @ry Would you approve this PR if I put the `zsh` and `ksh` tests into an if-CI block?

MarkTiedemann:
> If it runs under sh - it’s fine

I disagree.

Here is a counter-example: [https://travis-ci.com/denoland/deno_install/jobs/171474453](https://travis-ci.com/denoland/deno_install/jobs/171474453)

On `osx`:

```
$ set -o pipefail
```

On `linux`:

```
$ set -o pipefail
set: Illegal option -o pipefail
```

> I think this is being more complicated than necessary.

Maybe.

At the moment, I'm leaning towards removing `zsh` and `ksh` tests just so the other improvements from this PR can be merged.

MarkTiedemann:
I changed the scope to just re-add the `osx` tests.

Is this fine with you now, @ry?

.travis.yml
vladimyr: > I know that `sh` is the lowest common denominator. But again, `/bin/sh` is symlinking to `/bin/bash` on Mac and to `/bin/dash` on Ubuntu. So in that sense there's no "if it works under `sh`, then it'll work everywhere". If you follow the symlinks, it only works under `bash` or `dash` because those are completely different shells. For example, `set -o pipefail` works in Bash, but not in Dash. That's why the CI was failing for me a few times. If we would have shipped that, we would have broken the installation for Ubuntu users or people using the script in Linux jobs on Travis. `shellcheck` did _not_ warn about `set -o pipefail` being a bashism so linting didn't help either.

For sake of technical correctness `/bin/sh` is not symlinked on mac. Instead you have dedicated _proxy_ binary that invokes `/bin/bash` with `--posix` flag thus limiting its feature set to emulate POSIX shell aka `sh`. As you noticed that downgrade is not perfect same way `dash` might not be equally restrictive as `sh` would be. Bottom line removing testing inside actual shells **is a mistake**.
Take [`nvm`](https://github.com/creationix/nvm) for example (shell script itself). Their installer is `bash` only so they test only in `bash` but all other parts are tested across wide variety of common shells user might use: https://github.com/creationix/nvm/blob/master/.travis.yml

vladimyr:
> I think this is being more complicated than necessary. If it runs under sh - it’s fine

Correct me if I'm wrong but if we don't count BSDs `sh` became abstract entity:
1. macOS - uses proxy binary that invokes `bash` under the hood
2. Ubuntu - symlinked to `dash`
3. Arch - symlinked to `bash`
4. ...

Saying _it works in `sh`_ became almost like that _it works on my machine_ badge 🙃 
Regarding simplicity, ofc that end goal should be simplest possible **correct** solution. In this case we are trading correctness for simplicity? That doesn't sound ok?

PR #36, opened 2019-01-25
fix IHTMLDocument2_write not exist

### LINKED ISSUE #35: does not contain a method named 'IHTMLDocument2_write'
```
PS C:\Users\zhang> iex (iwr https://deno.land/x/install/install.ps1)
iex : Method invocation failed because [System.__ComObject] does not contain a method named 'IHTMLDocument2_write'.
At line:1 char:1
+ iex (iwr https://deno.land/x/install/install.ps1)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (IHTMLDocument2_write:String) [Invoke-Expression], RuntimeException
    + FullyQualifiedErrorId : MethodNotFound,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

zhmushan:
fix #35

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -51,7 +51,11 @@ if (!$Version) {
   if ($PSVersionTable.PSVersion.Major -lt 6) {
     $Response = Invoke-WebRequest 'https://github.com/denoland/deno/releases'
     $HTMLFile = New-Object -Com HTMLFile
-    $HTMLFile.write([System.Text.Encoding]::Unicode.GetBytes($Response.Content))
+    if ($HTMLFile.IHTMLDocument2_write) {
+      $HTMLFile.IHTMLDocument2_write($Response.Content)
+    } else {
+      $HTMLFile.write([System.Text.Encoding]::Unicode.GetBytes($Response.Content))
+    }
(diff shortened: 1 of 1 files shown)


ry:
cc @MarkTiedemann

MarkTiedemann:
LGTM.

I'd even replace the old line instead of using the try-catch.

`$HTMLFile.write([System.Text.Encoding]::Unicode.GetBytes($Response.Content))` is much cleaner to me.

zhmushan:
@MarkTiedemann
Done.

zhmushan:
ci is red.
I think it caused by a different powershell version, because the test passed on my machine.
Maybe trycatch should be retained?

MarkTiedemann:
I'd prefer to have an if-else based on the PowerShell version, or an if-else based on whether the IHTMLDocument2_write member exists on the ComObject. If we can't get that to work reliably, then try-catch.

MarkTiedemann:
Perfect, thanks!

PR #43, opened 2019-02-16
Remove Python installer

### LINKED ISSUE #41: Remove python installer
Now that we have a shell one, which has fewer deps than the python, and now that all the documentation has been updated to point at install.sh, we should remove the python installer.

MarkTiedemann:
Closes #41.

Should be merged on the 1st of March.


MarkTiedemann:
ping @ry 

It's the 1st of March.

PR #44, opened 2019-02-16
Clean-up CI

MarkTiedemann:
- Removed the Python CI.
- Moved Shell and PowerShell tests to the same job to improve CI speed.

In total, this PR removes 2 AppVeyor jobs and 4 Travis jobs.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,25 +1,4 @@
-environment:
-  matrix:
-    - TYPE: powershell
-    - TYPE: python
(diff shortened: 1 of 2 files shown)


ry:
While we still have python in the repo we should keep it in CI

MarkTiedemann:
But since the python code is deprecated, it should not be updated anymore (only removed), and if it's not updated anymore, do we really still need to test it in CI?

ry:
It doesn't hurt to keep it. We'll remove in march.

Issue #56, opened 2019-04-10
TLS 1.0 support


MarkTiedemann:
`https://deno.land/x/install/install.ps1` redirects to `https://raw.githubusercontent.com/denoland/deno_install/master/install.ps1`:

```
$ curl -I https://deno.land/x/install/install.ps1 | grep Location:
Location: https://raw.githubusercontent.com/denoland/deno_install/master/install.ps1
```

`deno.land` does **not** work with TLS 1.0.

```diff
$ curl --tls-max 1.2 https://deno.land/x/install/install.ps1
$ curl --tls-max 1.0 https://deno.land/x/install/install.ps1
-curl: (35) error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure
```

However, `raw.githubusercontent.com` works fine with TLS 1.0.

```
$ curl --tls-max 1.2 https://raw.githubusercontent.com/denoland/deno_install/master/install.ps1
$ curl --tls-max 1.0 https://raw.githubusercontent.com/denoland/deno_install/master/install.ps1
```

There have been a number of issues because PowerShell <v6 uses TLS 1.0 by default.

For example:

- https://github.com/denoland/deno_install/issues/51
- https://github.com/denoland/deno_install/issues/16
- https://github.com/denoland/deno/issues/1729

The default PowerShell version for Windows 8 is v4, for Windows 10 it is v5. So this impacts nearly all PowerShell users.

The following fixes are possible:

- (1) Changing the one-line install command to use TLS 1.2

```powershell
[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; iwr https://deno.land/x/install/install.ps1 | iex
```

- (2) Changing the one-line install command to point directly to GitHub

```powershell
iwr https://raw.githubusercontent.com/denoland/deno_install/master/install.ps1 | iex
```

- (3) Enabling TLS 1.0 for `deno.land`

cc @ry

hayd:
I believe (3) is a CloudFront setting https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValues-security-policy

but (1) also seems fine.

MarkTiedemann:
I'd like to avoid (1) purely because it makes the one-line install command very long...

MarkTiedemann:
@ry Any chance we can enable TLS 1.0 for `deno.land` in CloudFront? This is still an issue.

```diff
[...]

PR #57, opened 2019-04-10
ps: Add -UseBasicParsing flag

MarkTiedemann:
Closes: https://github.com/denoland/deno/issues/1965

Not sure whether we should also update the one-line install command: `iwr https://deno.land/x/install/install.ps1 -useb | iex`. I don't think this is a common issue, so only documenting this in the Known Issues section seems appropriate to me for now.

cc @ry

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,8 @@ Set-ExecutionPolicy RemoteSigned -Scope CurrentUser -Force
 ### Response content cannot be parsed
 
 ```
-PS C:\> iwr : The response content cannot be parsed because the Internet Explorer engine is not available, or Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again.
+PS C:\> iwr https://deno.land/x/install/install.ps1 | iex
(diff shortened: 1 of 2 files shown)


ry:
Is there a reason not to update the one-liner?

MarkTiedemann:
> Is there a reason not to update the one-liner?

Other than that this issue is not too common (because I think most devs start IE at least once to google "download Chrome"...), no.

ry:
@MarkTiedemann if there's no downside, then I'd advocate adding it so the docs can be simpler.

MarkTiedemann:
Right. Sounds good to me. I'll add `-useb`.

I'll also do another PR to change `-out` to `-outf` for the specific version installation (so we get PowerShell <v5 support there).

As well as update the docs in `denoland/deno` and update the install command in `denoland/deno_std`.

By the way, did you have any chance to look at https://github.com/denoland/deno_install/issues/56 yet? I think that's a more pressing issue at the moment since it affects nearly all Windows users.

MarkTiedemann:
@ry Added `-useb` for the latest install and changed `-out` to `-outf` for the specific install. Is this fine with you?

PR #62, opened 2019-05-23
Add brew install deno to README

hayd:


ry:
We should also update the manual https://deno.land/manual.html#binaryinstall

MarkTiedemann:
If someone updates the manual, please also recommend `scoop install deno` on Windows.

PR #75, opened 2019-08-19
Use DENO_INSTALL instead of DENO_DIR

nayeemrmn:
The install script is the only place in denoland where the cache directory variable `$DENO_DIR` is also used to determine Deno's installation directory. This separates the two into `$DENO_DIR` and `$DENO_HOME`.

Now `$DENO_HOME` can be set to a custom location in your profile (for updates or third party tools) without the cache defaulting to the same, potentially root-owned, location.

cc @ry

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ scoop install deno
 brew install deno
 ```
 
+## Environment Variables
+
+- `DENO_INSTALL` - The directory in which to install Deno. This defaults to `$HOME/.deno`. Not yet supported in the PowerShell installer ([#76](https://github.com/denoland/deno_install/issues/76)).
+
 ## Compatibility
 
 - The Shell installer can be used on Windows via the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).
--- a/README.md
+++ b/README.md
@@ -50,7 +50,9 @@ brew install deno
 
 ## Environment Variables
 
-- `DENO_INSTALL` - The directory in which to install Deno. This defaults to `$HOME/.deno`. Not yet supported in the PowerShell installer ([#76](https://github.com/denoland/deno_install/issues/76)).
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -31,8 +31,8 @@ else
 	deno_uri="https://github.com/denoland/deno/releases/download/${1}/deno_${os}_x64.gz"
 fi
 
-deno_dir=${DENO_DIR:-$HOME/.deno}
-bin_dir="${deno_dir}/bin"
+deno_home=${DENO_HOME:-$HOME/.deno}
```
j-f1: Should this fall back to `DENO_DIR` if `DENO_HOME` is not specified?

nayeemrmn: @j-f1 Well the point is draw a line between the two i.e. `DENO_HOME` is exclusively the installation directory, `DENO_DIR` is the cache.

For backwards compatibility? That would work for my use case, which is to set `DENO_HOME` to /opt/deno in my profile and generally let `DENO_DIR` default, but I think it would muddy the meanings of these variables. I'm not sure how many of whatever users there are at the moment are doing custom directory installations like this anyway.

ry:
Can we call it DENO_INSTALL instead of DENO_HOME (which I think is confusingly similar to DENO_DIR)?

Can you update the README too?

nayeemrmn:
@ry Done 👍. I also like `DENO_INSTALL` because it better reflects its specificity to this repo.

README.md
```diff
@@ -48,6 +48,12 @@ scoop install deno
 brew install deno
 ```
 
+## Environment Variables
+
+- `DENO_INSTALL` - The directory in which to install Deno. This defaults to `$HOME/.deno`.
+  Not yet supported in the PowerShell installer (#76).
```
ry: Can you give a concrete example of usage here?

nayeemrmn: 👍

Issue #82, opened 2019-10-28
Migrate to GitHub Actions


imbsky:

imbsky:
@ry I'm ready to submit the PR, what do you think?

ry:
Thanks but I'd rather merge this repo into the main repo. I don't see a purpose of having it separate anymore.

imbsky:
Yes, that's right. I have already written the config in https://github.com/imbsky/deno_install/commit/8838815794b19130d84b9a03e9969ccbef37f44e, so feel free to use them if you need.

imbsky:
CI is faster because the installation and execution of ShellCheck, shfmt, and PSScriptAnalyzer are completely separate from the test scripts and run in parallel.

PR #83, opened 2019-10-29
Use GitHub Actions

imbsky:
Closes: #82 

I understand that you plan to move the contents of this repository to the main repository, but probably these changes will be reusable even if you move them to the main repository. CI is faster than before because the installation and execution of ShellCheck, shfmt, and PSScriptAnalyzer are completely separate from the test scripts and run in parallel.


ry:
Ok I will use it- thanks. Clean up and landing it in #84

PR #89, opened 2019-12-06
Default DENO_INSTALL to $HOME/.local

### LINKED ISSUE #40: Change installation directory
Since https://github.com/denoland/deno/commit/e782ba1a6fd8342cddd961da9274dbd0244685d1  deno now stores cache files in platform dependent cache directories like `$HOME/Library/Caches` (mac).

Downloading the deno binary into `$HOME/.deno/bin` seems now inappropriate. 

Where should it be installed?

cc @flying-sheep

nayeemrmn:
Closes #40.

~To make `DENO_INSTALL` compatible with `XDG_BIN_HOME`, it now points to the binary directory directly rather than having `bin` appended to it. i.e. Installing to `/usr/local/bin` required `DENO_INSTALL=/usr/local`, now it would be `DENO_INSTALL=/usr/local/bin`.~

cc @ry

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -31,12 +31,12 @@ else
 	deno_uri="https://github.com/denoland/deno/releases/download/${1}/deno_${os}_x64.gz"
 fi
 
-deno_install=${DENO_INSTALL:-$HOME/.deno}
-bin_dir="${deno_install}/bin"
-exe="$bin_dir/deno"
+xdg_bin_home=${XDG_BIN_HOME:-$HOME/.local/bin}
+deno_install=${DENO_INSTALL:-$xdg_bin_home}
+exe="$deno_install/deno"
 
-if [ ! -d "$bin_dir" ]; then
-	mkdir -p "$bin_dir"
+if [ ! -d "$deno_install" ]; then
+	mkdir -p "$deno_install"
 fi
 
 curl -fL# -o "$exe.gz" "$deno_uri"
@@ -49,6 +49,6 @@ if command -v deno >/dev/null; then
 else
 	echo "Manually add the directory to your \$HOME/.bash_profile (or similar)"
 	echo "  export DENO_INSTALL=\"$deno_install\""
-	echo "  export PATH=\"\$DENO_INSTALL/bin:\$PATH\""
+	echo "  export PATH=\"\$DENO_INSTALL:\$PATH\""
 	echo "Run '$exe --help' to get started"
 fi
--- a/install.sh
+++ b/install.sh
@@ -31,12 +31,12 @@ else
 	deno_uri="https://github.com/denoland/deno/releases/download/${1}/deno_${os}_x64.gz"
 fi
 
-xdg_bin_home=${XDG_BIN_HOME:-$HOME/.local/bin}
-deno_install=${DENO_INSTALL:-$xdg_bin_home}
-exe="$deno_install/deno"
+deno_install=${DENO_INSTALL:-$HOME/.local}
+bin_dir=$deno_install/bin
+exe="$bin_dir/deno"
 
-if [ ! -d "$deno_install" ]; then
-	mkdir -p "$deno_install"
+if [ ! -d "$bin_dir" ]; then
+	mkdir -p "$bin_dir"
 fi
 
 curl -fL# -o "$exe.gz" "$deno_uri"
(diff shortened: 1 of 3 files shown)


flying-sheep:
Super cool, but this is the first time I hear of XDG_BIN_HOME existing. I can only find old patches, no primary source stating that it exists already. Do you have one for us?

nayeemrmn:
@flying-sheep Oh... looks like the sources I thought I had are those old patches. I saw this https://docs.rs/dirs/1.0.0/dirs/fn.executable_dir.html as a green light and didn't look around much more.

My spreading of misinformation aside, what do you think about using it anyway? `dirs` is used in Deno and it uses it. Otherwise I can skip the `XDG_BIN_HOME` check and just borrow its default for `DENO_INSTALL`.

flying-sheep:
Hmm, filed soc/dirs-rs#27. Let’s see, maybe we can get that stuff going again on freedesktop?

nayeemrmn:
Since `XDG_BIN_HOME` does not exist and there was a lot of contention towards adding it, I've removed it from this patch. Now it simply defaults to `$HOME/.local/bin` instead of `$HOME/.deno/bin`.

install.sh
```diff
@@ -31,8 +31,8 @@ else
 	deno_uri="https://github.com/denoland/deno/releases/download/${1}/deno_${os}_x64.gz"
 fi
 
-deno_install=${DENO_INSTALL:-$HOME/.deno}
-bin_dir="${deno_install}/bin"
+deno_install=${DENO_INSTALL:-$HOME/.local}
+bin_dir=$deno_install/bin
```
ry: Maybe leave the quotation marks? Without quotation marks maybe this breaks in the case that $deno_install has spaces in it

flying-sheep: This is a variable assignment. What you say is true for e.g. running a command, but variable assignments have their own rules:

At least in bash or zsh, assigning variables like this will work, AFAIK even with consecutive spaces.

So since the shebang is /bin/sh, if that links to bash or zsh it’ll work. Idk how it is for other POSIX shells.

rsp:
@nayeemrmn, will the default for `install.sh` be `$HOME/.local/bin` for both Mac and Linux (the readme now explicitly says just Linux) but the default for `install.ps1` will remain `$HOME/.deno/bin` on Windows, or will it be changed as well? (asking to make sure that the CI jobs keep working after the changes)

I'm asking here since you wrote it closes #40. Thanks.

nayeemrmn:
> @nayeemrmn, will the default for `install.sh` be `$HOME/.local/bin` for both Mac and Linux (the readme now explicitly says just Linux) but the default for `install.ps1` will remain `$HOME/.deno/bin` on Windows, or will it be changed as well? (asking to make sure that the CI jobs keep working after the changes)

@rsp It's for `install.sh` used on any platform. Windows uses the PowerShell installer and keeps its old default. I specified Linux in the README's description of the environment variable because `$HOME/.local/bin` is a Linux convention but for macOS... currently yes it uses the same default if installed with `install.sh` but I left it unspecified because I don't know if that's the correct location in the long term...

@ry Maybe the default should stay as `$HOME/.deno/bin` for both Windows _and_ macOS? That would make it effectively just `dirs::executable_dir().unwrap_or(home_dir.join(".deno")).join("bin")`. Actually, It would be really nice and conventional to just match the behaviour of this line going forward and stop caring.

rsp:
@nayeemrmn Thanks for the info. I hope the PowerShell installer don't change the default because it doesn't support `DENO_INSTALL` as far as I know (according to #76) so I need to rely on the hardcoded install path (currently `$HOME/.deno/bin`) to make it work on CI, like [here](https://github.com/rsp/deno-clipboard/blob/b89938ff/.github/workflows/ci.yml#L38-L39):
```
          iwr https://deno.land/x/install/install.ps1 -useb | iex
          echo ::add-path::$Home\\.deno\\bin
```
I fixed my CI config here rsp/deno-clipboard@b89938ff by using explicit `DENO_INSTALL` for Mac and Linux so it won't break next time when the defaults change again (and I see that it might, at least for Mac, from what you wrote), but the Windows build would break because it cannot be guarded by the `DENO_INSTALL`.

nayeemrmn:
@rsp This was reverted in #106.

PR #91, opened 2019-12-11
Add documents to manage multiple versions on Windows with Scoop

zhmushan:


README.md
```diff
@@ -68,6 +68,19 @@ asdf global deno 0.2.10
 asdf local deno 0.2.10
 ```
 
+**With [Scoop](https://scoop.sh):**
```
nayeemrmn: Duplicated link. Maybe link here https://github.com/lukesampson/scoop/wiki/Switching-Ruby-And-Python-Versions instead?

zhmushan: agree

MarkTiedemann:
I believe this PR is incorrect.

Switching `deno` versions using Scoop is not possible at the moment.

The linked article starts with the following command `scoop bucket add versions`. This adds an additional scoop bucket which contains major/minor versions of well-known programs.

The `main` bucket contains only `ruby` (https://github.com/ScoopInstaller/Main/blob/master/bucket/ruby.json). The `versions` bucket contains `ruby19` (https://github.com/ScoopInstaller/Versions/blob/master/bucket/ruby19.json) and `ruby24` (https://github.com/ScoopInstaller/Versions/blob/master/bucket/ruby24.json), for example.

As far as I know, that is the only concept of "versioning" that is currently possible with Scoop. There is no `scoop install deno@0.22.0`, as described in the README. Thus, I believe this PR should be reverted.

On another note, we can create our own bucket (for example, `https://github.com/denoland/deno-versions.git`) with a `bucket` directory and put `deno-v0-26-0.json`, `deno-v0-25-0.json`, etc. in there, and voilà, we got versioned Deno with Scoop. Users would then have to add the bucket via `scoop bucket add deno-versions https://github.com/denoland/deno-versions.git` and could then run, for example, `scoop install deno-v0-26-0`. (Not sure whether Scoop works with `.`'s in the tool name, perhaps even `deno-v0.26.0` could work.)

zhmushan:
image
@MarkTiedemann

MarkTiedemann:
@zhmushan Awesome! Thanks for correcting my incorrect correction attempt.

(Well, this didn't work when I started using Scoop a couple of years ago... Perhaps I should make a PR to update their documentation about this. :)

PR #92, opened 2020-01-18
Trivial installer messaging update on the post install.sh default shell profile filename path

jremi:
Added a minor/trivial case statement to check `$SHELL` environment variable and suggest the correct path based on the users actual running shell.

Reasoning is that many (specifically: macOS) users are running newer versions of macOS Catalina or higher which come shipped with `zsh` by default. 

(Ref: https://support.apple.com/en-us/HT208050)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,13 +48,8 @@ if command -v deno >/dev/null; then
 	echo "Run 'deno --help' to get started"
 else
 	case $SHELL in
-	/bin/zsh)
-		shell_profile=".zshrc"
-		break
-		;;
-	*)
-		shell_profile=".bash_profile"
-		;;
+	/bin/zsh) shell_profile=".zshrc" ;;
+	*) shell_profile=".bash_profile" ;;
 	esac
 	echo "Manually add the directory to your \$HOME/$shell_profile (or similar)"
 	echo "  export DENO_INSTALL=\"$deno_install\""
--- a/install.sh
+++ b/install.sh
@@ -47,8 +47,17 @@ echo "Deno was installed successfully to $exe"
 if command -v deno >/dev/null; then
 	echo "Run 'deno --help' to get started"
 else
-	echo "Manually add the directory to your \$HOME/.bash_profile (or similar)"
+	case $SHELL in
+	/bin/zsh)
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -47,8 +47,17 @@ echo "Deno was installed successfully to $exe"
...
-	echo "Manually add the directory to your \$HOME/.bash_profile (or similar)"
+	case $SHELL in
+	/bin/zsh)
+		shell_profile_filename=".zshrc"
+		break
+		;;
+	*)
+		shell_profile_filename=".bash_profile"
+		;;
+	esac
```
MarkTiedemann, line 58: Could you format the case statements like the case statements above? e.g.

```bash
case $(uname -s) in
Darwin) os="osx" ;;
*) os="linux" ;;
esac
```

I think `shell_profile_filename` could be shortened to `shell_profile`.

Otherwise, LGTM.

jremi: Hi,

The ci test runner forces `shfmt` (shell auto formatter) to make pretty.
https://github.com/denoland/deno_install/blob/master/.github/workflows/ci.yml

MarkTiedemann: Well, `shfmt` accepts both

```bash
case $SHELL in
/bin/zsh) shell_profile=".zshrc" ;;
*) shell_profile=".bash_profile" ;;
esac
```

```bash
case $SHELL in
/bin/zsh)
	shell_profile_filename=".zshrc"
	break
	;;
*)
	shell_profile_filename=".bash_profile"
	;;
esac
```
as "properly" formatted.

Again, I'd prefer the first formatting since it's similar to the formatting of the rest of the script.

jremi: @MarkTiedemann 

I just made 2 new commits to update the variable to `shell_profile` and also adjusted the case statement to use the similar formatting as rest of the script. 

We should be good to merge:

🙏

MarkTiedemann: Yep, thanks, @jremi!

MarkTiedemann:
LGTM.

MarkTiedemann:
cc @bartlomieju I think this PR is ready to be merged.

Issue #93, opened 2020-02-29
Installation with Homebrew doesn't work on Linux


Windvis:
The Deno website mentions the following installation option:
`Using Homebrew (macOS or Linux):`

But when I try to install it using `brew install deno` I see the following error message:

```
deno: macOS is required.
Error: An unsatisfied requirement failed this build.
```

Is the website outdated or should this work? :smile:

ry:
cc @chrmoritz

chrmoritz:
Refs: https://github.com/Homebrew/linuxbrew-core/commit/70eeef0981c404e149a6650fc63355e5090fa353

This is unfortunately related to the fact that `v8` (in `rust_v8`, a dependency of `deno`) still requires Python 2 to build. After the `python@2` removal (from earlier this year), there is just no reliable way for Linuxbrew anymore to ensure that the `python@2` build dependency is fulfilled, because not all Linux distribution are still shipping with a Python 2 version installed by default (unlike all current versions of macOS, but I'm not sure if this will still be the case after this fall).

In conclusion, we can't offer Linux support for installing `deno` with Homebrew anymore, until the Python 2 => 3 situation is sorted out and the Linux part should probably be removed from the docs until then.

ry:
Thanks for the update @chrmoritz 

Before anyone asks, the python3 migration is happening in chromium:
https://bugs.chromium.org/p/chromium/issues/list?q=label:Proj-Python3Migration
https://bugs.chromium.org/p/chromium/issues/detail?id=942720
There's nothing to be done in Deno until that project is complete.

Windvis:
Alright, thanks for explaining! Should I close this or should it be kept open until the docs are updated?

ry:
Let's leave it open until docs are updated.

clemens-tolboom:
In https://github.com/denoland/deno_install/pull/94#issuecomment-594035996

> Edit: deno 0.35 is now available via Homebrew on macOS. 🎉

Not sure this is the same for Linux

I checked @ MacOS
```
brew update
brew info deno
# deno: stable 0.35.0 (bottled)
```

chrmoritz:
It won't work on Linux, because of the missing Python 2 requirement. I'm currently experimenting if you can get it to work using PyPy2.7 instead.

PR #94, opened 2020-03-03
Indicate brew installs pre 0.35.x #93

clemens-tolboom:


chrmoritz:
That's not correct. The macOS package will be updated to 0.35 once the recent CI problem are resolved (already done as of now) and it's turn for `deno` in the currently large CI backlog.

It's just that you can't install it on Linux, because of the not satisfiable `python@2` requirement.

clemens-tolboom:
I was confused with the answer from `bartlomieju` in https://github.com/denoland/deno/issues/4047#issuecomment-593099880 ... closing

Issue #103, opened 2020-04-04
PowerShell installation instructions fail


Vandivier:
## repro steps and resulting error
1. open new PowerShell session as admin on Windows 10 (I do have ConEmu installed as a console wrapper)
2. run `iwr https://deno.land/x/install/install.ps1 -useb | iex` as instructed on Deno website
3. receive below error (text and image both presented)

![image](https://user-images.githubusercontent.com/5559355/78456160-0d504300-7670-11ea-98bf-b4e780f581ea.png)

Error text: "iex : Cannot validate argument on parameter 'Uri'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command again."

## other info
`cmd --version` results with `Version 10.0.18362.720`

ry:
Thanks for the report.

I'm not sure what the problem is - I don't have a windows machine to test immediately - but I believe another version of the command should work:
```
iwr https://deno.land/x/install/install.ps1 -useb -outf install.ps1; .\install.ps1 v0.39.0
```

Vandivier:
That version of the command does working, at least in PowerShell as admin after running `set-executionpolicy remotesigned`. Closing this issue now.

piscisaureus:
I also noticed that `iwr https://deno.land/x/install/install.ps1 -useb | iex` was broken earlier, but I just tried again and now it works fine. It looks like #104 did indeed fix the issue.

PR #104, opened 2020-04-04
fix: Not working install script for PS5

ErazerBrecht:
Incorrect github link was used for downloading `deno` when using 'non-core' powershell


ErazerBrecht:
Should fix #103

Issue #68, opened 2019-07-07
install.sh should work in windows mingw shell


ry:

BearD01001:
> Yes - somehow it’s not working in mingw denoland/deno_install#68
@BearD01001 try the powershell install command
(and delete the $HOME/.deno/bin/deno file)
>
>replied from @ry at [①deno/Lobby - Gitter](https://gitter.im/denolife/Lobby)

Leave it here in case of someone needed.

ccheng:
So I got it to work with Git Bash, which I think is actually mingw shell, and I've tested it on Mac, Windows, and Ubuntu.
Is it ok to make a pull request?  Not sure how to go about this...
Thanks!

ry:
@ccheng please do!

PR #107, opened 2020-04-14
make install.sh work in windows Git Bash / mingw shell

ccheng:
So I got install.sh to work with Git Bash, which I think is actually mingw shell, and I've tested it on Mac, Windows, and Ubuntu. #68

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,7 +20,10 @@ if [ $# -eq 0 ]; then
 			command grep -o "/denoland/deno/releases/download/.*/deno-${target}\\.zip" |
 			command head -n 1
 	)
-	if [ ! "$deno_asset_path" ]; then exit 1; fi
+	if [ ! "$deno_asset_path" ]; then
+		echo "Error: deno_asset_path not found."
+		exit 1;
+	fi
 	deno_uri="https://github.com${deno_asset_path}"
 else
 	deno_uri="https://github.com/denoland/deno/releases/download/${1}/deno-${target}.zip"
--- a/install.sh
+++ b/install.sh
@@ -23,7 +23,7 @@ if [ $# -eq 0 ]; then
 			command head -n 1
 	)
 	if [ ! "$deno_asset_path" ]; then
-		echo "Error: deno_asset_path not found."
+		echo "Error: Unable to find latest Deno release on GitHub."
 		exit 1;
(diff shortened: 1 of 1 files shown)


ccheng:
Not sure why ci / test (ubuntu-latest) failed.  Got to:
```
Deno was installed successfully to /home/runner/.deno/bin/deno
```
But then
```
./install_test.sh: 17: ./install_test.sh: ./install.sh: Permission denied
##[error]Process completed with exit code 126.
```

install.sh
```diff
@@ -4,9 +4,11 @@
 
 set -e
 
+win_target="x86_64-pc-windows-msvc"
 case $(uname -s) in
 Darwin) target="x86_64-apple-darwin" ;;
-*) target="x86_64-unknown-linux-gnu" ;;
+Linux) target="x86_64-unknown-linux-gnu" ;;
+*) target=$win_target ;;
```
MarkTiedemann, line 11: I don't think the default target should be Windows. Wikipedia has a long list of `uname -s` output: https://en.wikipedia.org/wiki/Uname For Linux, there's `FreeBSD`, `GNU`, `DragonFly`. For Windows, Wine uses `Windows_NT`, Cygwin uses `CYGWIN_NT-$version`, MingW uses `MINGW$arch_NT-$version`, MSYS uses `MSYS_NT-$version`.

Maybe `$OS = "Windows_NT"` would be a good check as well? (see: https://en.wikipedia.org/wiki/Environment_variable#Windows)

install.sh
```diff
@@ -20,15 +22,22 @@ if [ $# -eq 0 ]; then
 			command grep -o "/denoland/deno/releases/download/.*/deno-${target}\\.zip" |
 			command head -n 1
 	)
-	if [ ! "$deno_asset_path" ]; then exit 1; fi
+	if [ ! "$deno_asset_path" ]; then
+		echo "Error: deno_asset_path not found."
```
MarkTiedemann, line 26:
```suggestion
		echo "Error: Unable to find latest GitHub release."
```

install.sh
ccheng: Would it be okay to check for `$OS = "Windows_NT"` then if not Darwin, then default to Linux?

PR #113, opened 2020-05-14
Change DENO_INSTALL to DENO_INSTALL_ROOT

jmcantrell:
According to the docs (https://deno.land/manual/tools/script_installer), the environment variable that controls where deno is installed is `DENO_INSTALL_ROOT`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,21 +83,21 @@ scoop reset deno
 
 ## Environment Variables
 
-- `DENO_INSTALL` - The directory in which to install Deno. This defaults to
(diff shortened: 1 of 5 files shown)


MarkTiedemann:
@jmcantrell I think this is a misunderstanding:

`DENO_INSTALL_ROOT` is the path where Deno scripts will be installed if you `deno install` them.

`DENO_INSTALL` is the path where Deno itself will be installed.

jmcantrell:
Ok, thanks for the clarification.

PR #119, opened 2020-05-17
Updat the deno version to be installed with asdf

manuraj17:
Changes:
- In the README, update the deno version to be installed with asdf to 1.0.0

Although the change is trivial and not any sort of deal breaker, felt it's good to have the docs updated. I can update the whole doc if the version is available in every other installation source. I have only tried asdf, so I am not aware of other sources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,23 +59,23 @@ choco install deno
 ```sh
 asdf plugin-add deno https://github.com/asdf-community/asdf-deno.git
 
-asdf install deno 0.38.0
+asdf install deno 1.0.0
 
 # Activate globally with:
-asdf global deno 0.38.0
+asdf global deno 1.0.0
 
 # Activate locally in the current folder with:
-asdf local deno 0.38.0
+asdf local deno 1.0.0
 ```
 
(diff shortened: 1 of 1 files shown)


MarkTiedemann:
The reason why you find `0.38.0` in the docs rather than `1.0.0` is that this is for installing a _specific version_ rather than the _latest version_ of Deno.

In general, I think it's fine to update in this section: https://github.com/denoland/deno_install#install-and-manage-multiple-versions

But this section https://github.com/denoland/deno_install#install-specific-version should not be updated.

manuraj17:
@MarkTiedemann Ohkay. Gotcha. I have updated only the ASDF version. I will update the scoop version as well. I didn't want the document to give the feeling that it was outdated. Shall I update the PR with updating scoop version to point to 1.0.0?

MarkTiedemann:
@manuraj17 Sure, sounds good to me. :)

Deno `1.0.0` is available for Scoop: https://github.com/ScoopInstaller/Main/blob/master/bucket/deno.json

manuraj17:
@MarkTiedemann Ah okay cool! I was searching the scoop repo, I don't use that tool so was trying to find where it is. Let me update the PR and push.

MarkTiedemann:
LGTM.

cc @ry

Issue #117, opened 2020-05-16
Install.sh fails if unzip is not installed


NLstn:
If unzip is not installed, the install.sh should install it for you

nayeemrmn:
No, but it should be documented in the README.

MarkTiedemann:
Yep, this belongs in the "Known Issues" section of the README: https://github.com/denoland/deno_install#known-issues

hardy613:
Hello, 
I ran into this issue installing deno inside docker, I believe we could use `tar` to make the install more portable. As I was just working on this I don't mind making a PR.

Yamboy1:
I think part of this is because deno moved from a gzipped binary to a zip file. I remember having to update my aur package, and had to add unzip because I didn't realise that it wasn't part of base-devel (in arch) like gzip was. Can i ask why we removed that gzipped file? It was nice and easy to use imo

hardy613:
We have updated the readme in the `known issues` section and moved program checks to the start of the file. 

@MarkTiedemann I believe it is reasonable to close this issue.

PR #122, opened 2020-05-17
[issue-117] Unzip

hardy613:
## Description
Add unzip requirement to known issues in readme and add check for unzip in install file

we also moved all program checks at the start of the file, thanks @MarkTiedemann for reviews and suggestions!

## Issue
#117

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,8 @@ scoop reset deno
 ## Environment Variables
 
 - `DENO_INSTALL` - The directory in which to install Deno. This defaults to
-  `$HOME/.local` on Linux/macOS and `$HOME/.deno` on Windows. The executable
-  is placed in `$DENO_INSTALL/bin`. One application of this is a system-wide
-  installation:
+  `$HOME/.deno`. The executable is placed in `$DENO_INSTALL/bin`. One
+  application of this is a system-wide installation:
 
   **With Shell (`/usr/local`):**
 
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ scoop reset deno
 - The Shell installer can be used on Windows via the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).
 
 ## Known Issues
+- `unzip` is a requirement, `apt-get install unzip -y`
 
 ### Running scripts is disabled
 
--- a/README.md
+++ b/README.md
@@ -107,11 +107,13 @@ scoop reset deno
 
 ## Known Issues
 
-### The program [`unzip`](https://linux.die.net/man/1/unzip) is a requirement for the Shell installer
+### unzip is required
+The program [`unzip`](https://linux.die.net/man/1/unzip) is a requirement for the Shell installer
+
 
 ```sh
 $ curl -fsSL https://deno.land/x/install/install.sh | sh
-unzip is required to install deno
+unzip is required to install deno - https://github.com/denoland/deno_install#unzip-is-required
 ```
 
 **When does this issue occur?**
--- a/README.md
+++ b/README.md
@@ -111,8 +111,7 @@ scoop reset deno
 
 ```sh
 $ curl -fsSL https://deno.land/x/install/install.sh | sh
-######################################################################## 100.0%##O#- #
-sh: 39: unzip: not found
+unzip is required to install deno
 ```
 
 #### When does this occur
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ unzip is required to install deno
 ```
 
 #### When does this occur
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -106,6 +106,7 @@ scoop reset deno
 - The Shell installer can be used on Windows via the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).
 
 ## Known Issues
+- The program `unzip` is a requirement for the deno install process
```
MarkTiedemann, line 109: This is only true for the Shell installer, not for the PowerShell installer. Can you please adjust the wording accordingly?

MarkTiedemann:
@hardy613 I think it would be good to give some more details as well.

For example, the other known issue, ["Running scripts is disabled"](https://github.com/denoland/deno_install#running-scripts-is-disabled), (1) shows the output of a failed command, (2) answers the question "When does this issue occur?", and (3) answers the question "How can this issue be fixed?".

hardy613:
Sounds good

README.md
```diff
@@ -107,6 +107,20 @@ scoop reset deno
 
 ## Known Issues
 
+### The program [`unzip`](https://linux.die.net/man/1/unzip) is a requirement for the Shell installer
```
MarkTiedemann, line 110: Maybe shorten the issue heading to "unzip is required" and move the more detailed explanation into a text block below the heading?

Then you can also link to the known issue from the shell script (https://github.com/denoland/deno_install#unzip-is-required) rather than just printing `unzip is required to install deno`.

What do you think?

README.md
```diff
@@ -107,6 +107,20 @@ scoop reset deno
...
+
+```sh
+$ curl -fsSL https://deno.land/x/install/install.sh | sh
+unzip is required to install deno
+```
+
+#### When does this occur
+During the `instal.sh` process, `unzip` is used to extract the zip archive.
```
MarkTiedemann, line 118: nit

```suggestion
During the `install.sh` process, `unzip` is used to extract the zip archive.
```

README.md
```diff
@@ -107,6 +107,20 @@ scoop reset deno
...
 ## Known Issues
 
+### The program [`unzip`](https://linux.die.net/man/1/unzip) is a requirement for the Shell installer
+
+```sh
+$ curl -fsSL https://deno.land/x/install/install.sh | sh
+unzip is required to install deno
+```
+
+#### When does this occur
```
MarkTiedemann, line 117: nit

```suggestion
### When does this occur?
```

README.md
```diff
@@ -107,6 +107,20 @@ scoop reset deno
...
+During the `instal.sh` process, `unzip` is used to extract the zip archive.
+
+#### How can this be fixed
```
MarkTiedemann, line 120: nit

```suggestion
**How can this be fixed?**
```

README.md
hardy613: No problem, thanks for the reviews.

MarkTiedemann:
LGTM.

cc @ry

bartlomieju:
@MarkTiedemann are you still ok with these changes?

MarkTiedemann:
@bartlomieju This PR currently re-adds the "Running scripts is disabled" known issue, which I think may be a merging error.

I also think that the following check is not necessary since the script was, most likely, already downloaded using `curl`:

```shell
if ! command -v curl >/dev/null; then
	echo "Error: curl is required to install Deno." 1>&2
	exit 1
fi
```

But I think, if those two issues are fixed, it should be merged, yes.

bartlomieju:
> @bartlomieju This PR currently re-adds the "Running scripts is disabled" known issue, which I think may be a merging error.

I'm a bit lost, can you mark this bit in a diff?

README.md
```diff
@@ -110,3 +110,46 @@ scoop reset deno
...
+
+**When does this issue occur?**
+
+During the `install.sh` process, `unzip` is used to extract the zip archive.
+
+**How can this issue be fixed?**
+
+You can install unzip via `brew install unzip` on MacOS or `apt-get install unzip -y` on Linux.
+
+### Running scripts is disabled
```
MarkTiedemann, line 133: @bartlomieju This is no longer a known issue:

```
### Running scripts is disabled
...
```

Thanks to https://github.com/denoland/deno_install/pull/133, it is fixed.

#133 removed the temporary file for installing a specific version using the PS installer. Running a temp file could lead to the following issue: `File C:\install.ps1 cannot be loaded because running scripts is disabled on this system.`

No temp file -> no "running scripts is disabled" anymore. :)

bartlomieju:
Thank you @hardy613!

hardy613:
Amazing, thanks everyone!

PR #124, opened 2020-05-18
chore: Adds asdf instructions for latest version

anshulguleria:
Added instructions to install latest version of deno instead of specific
version as that would be more common use case.

Removed deno plugin url in `plugin add` command as `deno` is available
in `asdf` plugin list


bartlomieju:
@anshulguleria thank you for opening the PR. Can you resolve the conflict and present installing both `1.0.0` version and latest version?

anshulguleria:
Sure. Thanks for looking into it. Will update.

kt3k (approved):
LGTM

PR #118, opened 2020-05-17
Misc cleanup

MarkTiedemann:
- Removed unused variables
- Cleaned up whitespace
- Fix linting warnings

PR #105, opened 2020-04-05
Restrict Set-ExecutionPolicy scope

mattcbaker:
Restrict the scope on the recommended fix for the Powershell execution error. The options for the -Scope flag are "LocalMachine" (default), "CurrentUser", "Process", "UserPolicy" or "MachinePolicy". "Process" is the only option that doesn't make a permanent change -- it **only** allows the install script to run in the current Powershell session.

More info on -Scope here https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.security/set-executionpolicy?view=powershell-7

PR #133, opened 2020-05-28
Remove temp file from PS install command; Remove PS linting

### LINKED ISSUE #114: Removing temporary file from PowerShell install command
Currently the command looks like this:

```powershell
iwr https://deno.land/x/install/install.ps1 -useb -outf install.ps1; .\install.ps1 v1.0.0
```

We save the installer as a temporary file because we have no way of passing the version into the installer while piping, in contrast to the Shell version:

```powershell
curl -fsSL https://deno.land/x/install/install.sh | sh -s v1.0.0
```

I propose to add a variable `v`. If the variable is set, the `install.ps1` script will assume it's the Deno version that the user wants to install. If not, the script will work just as it works now.

So this method is similar to the "install latest" command:

```powershell
iwr https://deno.land/x/install/install.ps1 -useb | iex
```

Only difference is the added variable assigment beforehand:

```powershell
$v="v1.0.0"; iwr https://deno.land/x/install/install.ps1 -useb | iex
```

Example that demonstrates that the variable will be available to the invoked script:

```
> type v.ps1
write-host v=$v
> $v="v1.0.0"; type v.ps1 | iex
v=v1.0.0
```

I see two advantages:

1. The command is shorter.
2. We don't create a temporary file.

cc @ry

MarkTiedemann:


MarkTiedemann:
To test that this is working end-to-end, you can run: `$v="v1.0.0"; iwr https://raw.githubusercontent.com/MarkTiedemann/deno_install/fix-114/install.ps1 -useb | iex`.

MarkTiedemann:
Since this PR removes the "Known Issues" section, it makes the following PR obsolete: https://github.com/denoland/deno_install/pull/105

MarkTiedemann:
Linting fails because the cleanup from https://github.com/denoland/deno_install/pull/118/ is not merged yet.

MarkTiedemann:
After a bit of CI debugging, this seems like another GitHub Actions bug...

https://github.com/denoland/deno_install/pull/133/checks?check_run_id=718693105

```
DEBUG:   38+  >>>> Write-Host $LASTEXITCODE
0
DEBUG:    3+ if ( >>>> (Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE }
DEBUG:    3+ if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) {  >>>> exit $LASTEXITCODE }
##[error]Process completed with exit code 4.
```

`$LASTEXITCODE` is defined and has a value of `0`, as can be seen by the output of the `Write-Host $LASTEXITCODE` command.

Running `Test-Path -LiteralPath variable:\LASTEXITCODE`, however, which should return `True` if the variable `$LASTEXITCODE` is defined and `False` if it is undefined, does error out and produces a `$LASTEXITCODE` of `4` which means "_Cannot Access File_".

So I think there may be a GitHub Actions issue here where `variable:\LASTEXITCODE` is checking for the file `LASTEXITCODE` on the drive `variable` rather than the existence of the variable `$LASTEXITCODE`... Or something along those lines.

Not sure why this issue is showing up now, only in this PR, though. This is weird.

MarkTiedemann:
Now it's exit code 1. Then exit code 3.

```
DEBUG:   39+  >>>> Write-Host $LASTEXITCODE
0
DEBUG:   40+  >>>> $LASTEXITCODE = 0
DEBUG:   41+  >>>> Write-Host $LASTEXITCODE
0
DEBUG:   42+  >>>> exit $LASTEXITCODE
DEBUG:    3+ if ( >>>> (Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE }
DEBUG:    3+ if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) {  >>>> exit $LASTEXITCODE }


##[error]Process completed with exit code 3.
```

None of this makes any sense.

MarkTiedemann:
Well, I figured it out.

`PSScriptAnalyzer`, the linter, sets the exit code to the number of lint warnings + errors. There were 0 errors, and 1/3/4 warnings.

Not sure why this doesn't properly show up in `$LASTEXITCODE`.

Not sure why the `-ExcludeRule` did not work.

I'm considering to drop `PSScriptAnalyzer` (which I initially added to this repo) - it's showing 2 false positives for 2 short files at the moment anyways...

MarkTiedemann:
@bartlomieju This is ready to be merged now.

Scope has changed a bit - not just removing the temp file and the "Known Issues" section, but also removing the PS linter.

(I think PSScriptAnalyzer may be re-integrated later - similar to how shellcheck was disabled for a while, then re-integrated. In any case, it should be integrated with a fixed version...)

README.md
```diff
@@ -29,7 +29,7 @@ curl -fsSL https://deno.land/x/install/install.sh | sh -s v0.38.0
 **With PowerShell:**
 
 ```powershell
-iwr https://deno.land/x/install/install.ps1 -useb -outf install.ps1; .\install.ps1 v0.38.0
+$v="v0.38.0"; iwr https://deno.land/x/install/install.ps1 -useb | iex
```
bartlomieju, line 32: Maybe update to version >= 1.0.0?

MarkTiedemann: Yeah, I was thinking about that, too. However, there are multiple places where 0.38.0 is used, both in the tests and the readme. I think this should be a follow-up PR.

PR #145, opened 2020-06-01
Fix old PS temp file installation

### LINKED ISSUE #144: Update PowerShell specific-version installation instructions for GitHub Releases
While working on updating the PowerShell specific-version installation (https://github.com/denoland/deno_install/pull/143), I noticed that the releases page (https://github.com/denoland/deno/releases) uses the old temp file installation, e.g. `iwr $url -useb -outf install.ps1; .\install.ps1 $version` rather than the new one using the `$v` variable: `$v="$version"; iwr $url -useb | iex`. The temp file installation doesn't work anymore with the new installer. So I think we either have to update the releases with the new instructions or change the installer so the temp file installation works again. I think both are viable options so I'm looking for feedback on this topic.

cc @bartlomieju 

EDIT: In any case, it would make sense to merge (or close) #143 before updating the instructions.

MarkTiedemann:
Fixes #144.

cc @ry @bartlomieju

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -4,9 +4,16 @@
 
 $ErrorActionPreference = 'Stop'
 
+if ($v) {
+  $Version = "v${v}"
+}
+if ($args.Length -eq 1) {
+  $Version = $args.Get(0)
+}
+
 $DenoInstall = $env:DENO_INSTALL
 $BinDir = if ($DenoInstall) {
-    "$DenoInstall\bin"
+  "$DenoInstall\bin"
 } else {
   "$Home\.deno\bin"
 }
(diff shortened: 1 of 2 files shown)


install.ps1
```diff
@@ -4,9 +4,16 @@
...
+  $Version = "v${v}"
+}
+if ($args.Length -eq 1) {
+  $Version = $args.Get(0)
+}
+
 $DenoInstall = $env:DENO_INSTALL
 $BinDir = if ($DenoInstall) {
-    "$DenoInstall\bin"
+  "$DenoInstall\bin"
```
MarkTiedemann, line 16: Unrelated formatting.

install.ps1
```diff
@@ -4,9 +4,16 @@
 
 $ErrorActionPreference = 'Stop'
 
+if ($v) {
+  $Version = "v${v}"
+}
+if ($args.Length -eq 1) {
+  $Version = $args.Get(0)
```
MarkTiedemann, line 11: I gave the first argument precedence over $v since it's a more direct way to provide the version number to the script.

MarkTiedemann:
Sorry again about breaking the specific version installer on the releases page. I only checked the website and the manual for any references to the specific version installer, which both don't mention it. So I thought it was safe to only update the README in this repo.

Hope this fix will make up for my error earlier.

PR #148, opened 2020-06-03
Add support for shell installation on Windows

### LINKED ISSUE #68: install.sh should work in windows mingw shell

caspervonb:
This adds support for installing the Windows version of Deno from the
shell script which should work as long as the user has a basic UNIX like
environment (MSYS, WSL, MingW, et cetera).

It works for GitHub runners as-well which means there's less friction in
setting up a test matrix which installs Deno.

This resolves #68 and closes #107

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,16 @@ jobs:
           brew install shfmt shellcheck
           shfmt -d .
 
+      - name: shellcheck
+        if: matrix.os == 'windows-latest'
+        run: |
+          curl -LO https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.zip
+          mkdir ~/.shellcheck
+          unzip shellcheck-stable.zip -d ~/.shellcheck
+          mv ~/.shellcheck/shellcheck-stable.exe ~/.shellcheck/shellcheck.exe
+          echo "::add-path::$HOME/.shellcheck"
+
       - name: tests shell
-        if: matrix.os != 'windows-latest'
         shell: bash
         run: ./install_test.sh
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*       text=auto
+*.sh    text eol=lf
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ scoop reset deno
 
 ## Compatibility
 
-- The Shell installer can be used on Windows via the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).
+- The Shell installer can be used on Windows with [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about), [MSYS](https://www.msys2.org) or equivalent set of tools.
 
 ## Known Issues
 
--- a/install.sh
+++ b/install.sh
@@ -14,10 +14,14 @@ if ! command -v unzip >/dev/null; then
 	exit 1
 fi
 
-case $(uname -s) in
-Darwin) target="x86_64-apple-darwin" ;;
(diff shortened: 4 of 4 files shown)


caspervonb:
There are some line ending errors that I inadvertently fixed; reverted those changes in the latest force push.

install_test.sh
```diff
@@ -3,7 +3,12 @@
 set -e
 
 # Lint.
-shellcheck -s sh ./*.sh
+
+# FIXME (caspervonb) was unable to install spellcheck via choco on
+# windows-latest so only run spellcheck if available
+if [ -x "$(command -v spellcheck)" ]; then
```
MarkTiedemann, line 9: spellcheck != shellcheck

`shellcheck` is available for Windows.

caspervonb: Oops! indeed; will install **shell**check via choco in the workflow.

.github/workflows/ci.yml
```diff
@@ -20,8 +20,11 @@ jobs:
           brew install shfmt shellcheck
           shfmt -d .
 
+      - name: spellcheck
```
MarkTiedemann, line 23:
```suggestion
      - name: shellcheck
```

MarkTiedemann:
@caspervonb You could add a note to the Compatibility section of the Readme: https://github.com/denoland/deno_install#compatibility

.github/workflows/ci.yml
caspervonb: My brain is just really hung up on the word spellcheck this morning...

caspervonb:
> The Shell installer can be used on Windows with [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about), [MSYS](https://www.msys2.org) or equivalent set of tools.

This leaves WSL and MSYS as officially supported (altho; WSL is not tested here); anything else will most likely work but it's assumed that if you have assembled your own coreutils etc you know what you're doing.

MarkTiedemann:
https://github.com/denoland/deno_install/pull/148/checks?check_run_id=734138228

`choco` took almost 3 minutes to install `shellcheck.exe`. Not sure what's up with that. Is it always this slow?

If it is, can you perhaps try installing via `scoop`?

Otherwise, I guess we could also install it manually: `curl -LO https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.zip && powershell -c "Expand-Archive shellcheck-v0.7.1.zip -Destination ."`

> anything else will most likely work but it's assumed that if you have assembled your own coreutils etc you know what you're doing.

Yeah, I think that's fair.

> WSL is not tested here

PR is welcome. :)

caspervonb:
> choco took almost 3 minutes to install shellcheck.exe. Not sure what's up with that. Is it always this slow?

Not sure; I can't test locally as I only have Linux and MacOS; seems to take about two minutes on average.

> If it is, can you perhaps try installing via scoop?

It seems `scoop` is not available for GitHub workers based on [this list](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md)

> Otherwise, I guess we could also install it manually

Comes down to about 30 seconds to download from GitHub releases; unzip, alias, add to path, etc.

caspervonb:
Went ahead with using the binary release of shellcheck as it cuts the build time from more than two minutes to a few seconds.

In the long run; we should probably just move shellcheck into its own action to avoid per platform specific installation methods in the main workflow but that's a different concern.

nayeemrmn:
> > WSL is not tested here
> 
> PR is welcome. :)

It's nonsensical to go out of the way to support WSL. It's whole point is for the Linux option to "just work" (which it does here) and if it doesn't, that's a bug/limitation of WSL. We shouldn't test for it. This PR adds support for the other listed options.

PR #149, opened 2020-06-09
#6199 Deno Command not found

damiantoczek:
The added directories won't take effect until people run: source ~/$shell_profile

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,5 +59,7 @@ else
 	echo "Manually add the directory to your \$HOME/$shell_profile (or similar)"
 	echo "  export DENO_INSTALL=\"$deno_install\""
 	echo "  export PATH=\"\$DENO_INSTALL/bin:\$PATH\""
+	echo "After you have added the directories, run the following command:"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -59,5 +59,7 @@ else
 	echo "Manually add the directory to your \$HOME/$shell_profile (or similar)"
 	echo "  export DENO_INSTALL=\"$deno_install\""
 	echo "  export PATH=\"\$DENO_INSTALL/bin:\$PATH\""
+	echo "After you have added the directories, run the following command:"
```
MarkTiedemann, line 62:
```suggestion
	echo "After you have added the directories, reload your profile:"
```
Might be better to give a reason for the command.

damiantoczek: About the code, I guess it's not the best option because it's not persistent and it will only work for the current session inside your terminal. Where inserting it into the `.bashrc` is persistent.

ry:
Thanks for the patch @damiantoczek but I don't think these are necessary.

PR #150, opened 2020-06-10
Add support for Macports macOS package manager

elcheco:
There's other and older package manager for macOS (OS X) where deno has its maintained port

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,12 @@ scoop install deno
 brew install deno
 ```
 
+**With [Macports](https://ports.macports.org/port/deno/summary):**
+
+```sh
+sudo port install deno
+```
+
 **With [Chocolatey](https://chocolatey.org/packages/deno):**
 
 ```powershell


elcheco:
I am not, but I try to ask guys if they can do it. I tested it on my local, and it works, so there should not be a problem... I'll let you know

MarkTiedemann:
I found the Portfile here: https://github.com/macports/macports-ports/blob/master/devel/deno/Portfile It seems like it has to be updated manually.

elcheco:
> I found the Portfile here: https://github.com/macports/macports-ports/blob/master/devel/deno/Portfile It seems like it has to be updated manually.

Yes, but every port has its maintainer who is doing it after the new version is released.
It's basically coming from a similar FreeBSD "port" system.

hayd:
This does seem to be actively maintained: it's updated to 1.6.1 now.

ry (approved):
Thanks!

PR #152, opened 2020-06-24
Make install.ps1 cross-platform.

Summpot:


MarkTiedemann:
Hi @Summpot, thanks for your PR!

Unfortunately, I think it's a bit over-ambitious in scope:

1 - You switched from parsing the releases page to using the GitHub API for determining the latest version. That is not an option since the GitHub API is rate-limited, and users providing an API token is not an acceptable requirement for the installer.

2 - You also made the PowerShell installer cross-patform. This is actually something that we did previously (see https://github.com/denoland/deno_install/pull/25/), but that was removed since PowerShell Core is not a widely-used shell on Unix systems. The recommendation is Shell on Unix, PowerShell on Windows.

There is an open PR for making the Shell installer cross-platform: https://github.com/denoland/deno_install/pull/148 However, there's also the argument that we should not specifically support Shell on Windows since the Windows Subsystem for Linux is common now.

Personally, I'm not sure whether we should or should not make the installers cross-platform.

cc @ry - Any thoughts?

Summpot:
Reason? Because powershell supports cross-platform, and this script has no obvious label that it cannot cross-platform (I planned to use this script on linux at the beginning).
As for the rate limit of the Github API,
I think that normal users are unlikely to trigger this limit.

MarkTiedemann:
> As for the rate limit of the Github API, I think that normal users are unlikely to trigger this limit.

Normal users are very likely to trigger the limit. I have personally triggered the limit from both a company and a university network. Normal users also use CI services where shared servers are used that you have no control over. The CI of this repo has previously hit this rate limit in CI, before we moved to parsing the release page. Again, this is not an option.

> this script has no obvious label that it cannot cross-platform

This script is advertised on the https://deno.land homepage where it is very clearly labelled as Windows only: `PowerShell (Windows)` (or `Shell (Mac, Linux)` ).

> Reason?

Mostly maintenance and complexity.

Let me ask the other way around: Does your Linux distribution not come with some sort of POSIX shell out-of-the-box that you can use to install Deno? Why do you insist on using PowerShell on Linux?

Again, I have previously made the PowerShell installer cross-platform. I think it's a good idea in general. But all code comes with costs. This is something that needs to be carefully considered.

PR #161, opened 2020-09-19
Update install instructions for Arch Linux

eastack:


solson (approved):
LGTM - for anyone unfamiliar, AUR is the Arch User Repository for unofficial packages, while pacman is the official package manager.

daemon1024:
It seems like the `deno` package on pacman is on version `1.4.2` and `deno-bin` on AUR is version `1.4.6`. So it would be great if both of these versions are visible in the README.

MarkTiedemann:
> @rwrr It seems like the `deno` package on pacman is on version `1.4.2` and `deno-bin` on AUR is version `1.4.6`. So it would be great if both of these versions are visible in the README.

I don't think that's a good idea because then the README will have to be repeatedtly updated to show the latest version.

solson:
@MarkTiedemann That is not true, it just needs to explain both `yay -S deno-bin` (as before this PR) and `pacman -S deno` (instead of `yay -S deno`).

MarkTiedemann:
Oh, I misunderstood. I read @rwrr's comment as saying that both versions, pacman 1.4.2 and AUR 1.4.6, should be included in the README.

Yes, I'm fine with both pacman `deno`  and AUR `deno-bin` being included (as long as no specific version numbers are mentioned).

Nircek:
The month passed, and there is still only an AUR solution in the README.

UnlimitedCookies:
As an archlinux user I can approve the proposed changes. The official repositories are deemed trustworthy and thus should be the ones recommended here.  The Arch User Repository package `deno-bin`  has been removed in favor of the new official package, because it provides no real benefit. The only relevant AUR package at this point is [`deno-git`](https://aur.archlinux.org/packages/deno-git/), which downloads and builds directly from the git repo on the user's computer. If you want to mention this package (which is not mandatory imo), a note about it being user-provided content should be included.

eastack:
Should the AUR installation method be added back? The forked repository under my personal account has been deleted. If I modify the content, should I reopen a pull request?

kt3k (approved):
Sorry for the delay in review. This looks good to me. Thank you for the inputs & advise.

PR #163, opened 2020-10-15
Use GITHUB_PATH instead of add-path

peaceiris:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

> Action and workflow authors who are setting environment variables via stdout should update any usage of the set-env and add-path workflow commands to use the new environment files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           mkdir ~/.shellcheck
           unzip shellcheck-stable.zip -d ~/.shellcheck
           mv ~/.shellcheck/shellcheck-stable.exe ~/.shellcheck/shellcheck.exe
-          echo "::add-path::$HOME/.shellcheck"
+          echo "${HOME}/.shellcheck" >> ${GITHUB_PATH}
 
       - name: tests shell
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           mkdir ~/.shellcheck
           unzip shellcheck-stable.zip -d ~/.shellcheck
           mv ~/.shellcheck/shellcheck-stable.exe ~/.shellcheck/shellcheck.exe
-          echo "::add-path::$HOME/.shellcheck"
+          echo "${HOME}/.shellcheck" >> ${GITHUB_PATH}
```
MarkTiedemann, line 30: This code is running if `matrix.os == 'windows-latest'`. So this is PowerShell rather than Bash.

```suggestion
          echo "$HOME/.shellcheck" >> $env:GITHUB_PATH
```

MarkTiedemann (changes requested):
Thanks for the PR, @peaceiris!

peaceiris:
Updated!

bartlomieju (approved):
LGTM, thanks @peaceiris

PR #167, opened 2020-11-10
 feat: use direct mirror for latest release

### LINKED ISSUE #115: Clean version API
The _"Keep this script simple and easily auditable"_ `install.ps1` script has, at its core, a 20 line mess to determine the latest Deno version for Windows:

https://github.com/denoland/deno_install/blob/bd28456269a17a0809e7ca518e199bb9daad557b/install.ps1#L32-L52

(And I know it's a mess because I wrote the damn mess. :))

Deno being on v1.0.0 now, I think it's time to clean up a bit...

I know @lucacasonato tried to create a version API using a worker in https://github.com/denoland/deno_website2/pull/180.

I'd propose an even simpler API: All it's supposed to do is return the latest version released and available on GitHub Releases for each platform as a string, e.g. `GET /api/version/windows` returns `v1.0.0`.

Internally, the API would make the same mess that the installer script does at the moment, e.g. parse the HTML from `https://github.com/denoland/deno/releases`, check for the first anchor tag that matches a Windows release, return the number (plus some caching maybe).

If we had an API like that, those 20 lines would be a single one...

Not to mention that the same mess happens in the `install.sh` script which would benefit massively from this API as well.

maximousblk:
Can't list any major advantage over the earlier method but can say that this is the intended way of getting the latest from a repo and it also make it more auditable.

fixes #115

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -26,25 +26,7 @@ $Target = 'x86_64-pc-windows-msvc'
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 $DenoUri = if (!$Version) {
-  $Response = Invoke-WebRequest 'https://github.com/denoland/deno/releases' -UseBasicParsing
-  if ($PSVersionTable.PSEdition -eq 'Core') {
-    $Response.Links |
-      Where-Object { $_.href -like "/denoland/deno/releases/download/*/deno-${Target}.zip" } |
-      ForEach-Object { 'https://github.com' + $_.href } |
-      Select-Object -First 1
-  } else {
-    $HTMLFile = New-Object -Com HTMLFile
-    if ($HTMLFile.IHTMLDocument2_write) {
-      $HTMLFile.IHTMLDocument2_write($Response.Content)
-    } else {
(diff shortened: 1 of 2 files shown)


MarkTiedemann:
@maximousblk That's awesome! Thanks a lot!

maximousblk:
if it helps:

when you push a tag to release deno, the std and deno versions are added to the [`tags`](https://api.github.com/repos/denoland/deno/tags?per_page=100) but not [`releases`](https://api.github.com/repos/denoland/deno/releases?per_page=100) (both different) so for the time between pushing the tag, building the binaries, and creating a release, the assets also don't exist. but after the binaries are built and release is created, the assets are available.

Now the problem arises because, on the releases page of a repo, both tags and releases are listed so scraping the releases page isn't a good idea.

Yes, imo `deno upgrade` should also use this logic instead of scraping the releases page

lucacasonato:
> Is this still an issue? Is it possible for a release to exist without all assets?

No. CI creates draft releases which we only publish once all assets are attached.

> Yes, imo deno upgrade should also use this logic instead of scraping the releases page

It will with the patch that nightly @crowlKats is working on.

MarkTiedemann (approved):

ry (approved):
Thanks - LGTM

PR #157, opened 2020-07-31
ask to automatically add Deno to the PATH

SirJosh3917:
This modification to the script will ask the user if they want to automatically add Deno to their PATH.

PR #177, opened 2021-01-13
Add `Deno` to path

japrozs:
When you install deno on MacOS, you have to manually add Deno to path. with the change that I have proposed, Deno will add itself to the PATH to the files(like .bash_profile, .bashrc, .zshrc) which are loaded when any terminal is launched

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -38,6 +38,10 @@ unzip -d "$bin_dir" -o "$exe.zip"
 chmod +x "$exe"
 rm "$exe.zip"
 
+echo 'export PATH=$HOME/.deno/bin:$PATH' >>$HOME/.bash_profile
+echo 'export PATH=$HOME/.deno/bin:$PATH' >>$HOME/.zshrc
+echo 'export PATH=$HOME/.deno/bin:$PATH' >>$HOME/.bashrc
+
 echo "Deno was installed successfully to $exe"
 if command -v deno >/dev/null; then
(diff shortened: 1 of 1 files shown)


nayeemrmn:
I believe we don't do this because there is a lot of magic and presumption involved in automatic PATH configuration. For example: writing user directory profiles when the installation could be system wide; writing anything at all when one is just trying to download a copy of the binary.

The implementation here is improper, it creates and non-idempotently adds lines to multiple profiles.

Japroz-Saini:
This method is great, because if it was not able to add deno the the PATH it does nothing and asks the user to add it, but if it does add it to the PATH, do user doesn't have to do a thing

MarkTiedemann:
Thank you, @Japroz-Saini.

As with the previous PRs that asked to add this feature (for example, https://github.com/denoland/deno_install/pull/52
https://github.com/denoland/deno_install/pull/157), I believe this PR will be closed without merge.

The primary reason being that the installer should not mess with people's config files; a secondary reason being the complexity in adding to the `$PATH` on various platforms/shells.

PS: Your patch only adds to the files, but doesn't `source` them again.

caspervonb:
While I get the why; we're probably the few install scripts that don't add our bin path line to ~/.bashrc.

ry:
Thanks for the patch, but I'd prefer to keep the install script simple.

PR #178, opened 2021-01-26
chore: Automate set env for linux&macos

youngjuning:


kt3k:
duplicate of #177 ( #52, #157 )

PR #188, opened 2021-07-08
Replace `DENO_INSTALL` with `DENO_INSTALL_ROOT`

e-kwsm:
(denoland/deno#4787)

I’m not sure if this is intentional, but deno uses `DENO_INSTALL_ROOT`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,21 +115,21 @@ scoop reset deno
 
 ## Environment Variables
 
-- `DENO_INSTALL` - The directory in which to install Deno. This defaults to
-  `$HOME/.deno`. The executable is placed in `$DENO_INSTALL/bin`. One
+- `DENO_INSTALL_ROOT` - The directory in which to install Deno. This defaults to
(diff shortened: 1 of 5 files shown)


MarkTiedemann:
AFAIK, `DENO_INSTALL` is where the `deno` executable is installed. `DENO_INSTALL_ROOT` is where `deno install` installs Deno programs.

e-kwsm:
deno and deno_install respectively use `DENO_INSTALL_ROOT` and `DENO_INSTALL`, but the both default to `~/.deno`.

Is there necessity to distinguish them?

nayeemrmn:
Yes this is intentional. It's illogical for these to be bound to the same spot, other tools like `cargo` also have this distinction. For example, a system-wide deno installation with user-specific installs. If we _did_ unify them it would be the other way around.

e-kwsm:
Understood. Thank you for your comments.

PR #190, opened 2021-07-25
Add Canary Installer And Its Tests

koumaza:


MarkTiedemann:
Rather than copying and modifying the current installer scripts, I'd prefer to integrate canary installations into the existing scripts.

lucacasonato:
I would be in favor of not supporting this at all, and instead having users install a stable version and the `Deno upgrade` to canary. The install script should stay super simple and easily auditable.

MierenManz:
I agree with Luca on this one.
Canary can be quite buggy compared to stable.

koumaza:
> I would be in favor of not supporting this at all, and instead having users install a stable version and the `Deno upgrade` to canary. The install script should stay super simple and easily auditable.

`deno upgrade --canary` and `deno upgrade --canary --version ${HEXHASH}`, these are useful during the upgrade.
Nevertheless, if there is no installation script of Canary version, users who wants to install Canary from the beginning and
is it not costly for users who routinely install Canary?
Or, is there the alternative method already?

kt3k:
@koumaza For what situation do you need canary versions?

We sometimes use canary versions in CI. In that case we use [setup-deno](https://github.com/denoland/setup-deno) github action with `deno-version: canary`. Is this perhaps what you need?

caspervonb:
I'm in favour, if we add it as a named alias in the main install script.
Used to use this for CI (probably still do on some repos) and for services other than GitHub I can see it's still useful as a setup script.

To upgrade is extra conditional steps that don't fit into a test matrix nicely.

koumaza:
Hi @kt3k
> @koumaza For what situation do you need canary versions?
> 
> We sometimes use canary versions in CI. In that case we use [setup-deno](https://github.com/denoland/setup-deno) github action with `deno-version: canary`. Is this perhaps what you need?

I am necessary for normally use. (I'd think there are various reasons for using Canary.)
It is useful to be able to install Canary directly with POSIX Shell or Pwsh.
If who want to use Canary regularly, use a stable version of installation script, then use `deno upgrade --canary` to be a time loss.
Also, in CI, [setup-deno](https://github.com/denoland/setup-deno) is useful, but can only be used with GitHub Actions.

MarkTiedemann:
I'd be in favor of merging into the main script. For powershell, use a different variable name, e.g. `$c`. For shell, check if the first arg starts with `"v"` - if so, it's a version, otherwise the hash.

lucacasonato:
Thanks, but closing without merge as there isn't sufficient interest to warrant the increased complexity.

Issue #199, opened 2021-10-12
Exec format error on virtual Debian arm64


mathiasrw:
On a virtual debian (arm64) on a M1 CPU

I get 

    -bash: /home/debian/.deno/bin/deno: cannot execute binary file: Exec format error

when I 

    curl -fsSL https://deno.land/x/install/install.sh | sh
    deno -v

herkulessi:
Maybe even before downloading the AMD64 build?

PR #203, opened 2022-02-12
Fail on ARM/aarch architecture

herkulessi:
Script now failes on arm/aarch as architecture, closes issue #199


herkulessi:
NEVERMIND, this is bad, it also failes on M1 macs (which get a release, dunno if the script works for them tho....)

PR #209, opened 2022-05-13
Third party build support for Linux ARM64

### LINKED ISSUE #199: Exec format error on virtual Debian arm64

noxifoxi:
Enable installation of third party builds by passing the '-y' flag to the script otherwise block installation on ARM64 systems.

Not sure if third party builds are considered safe for the official installer, but I'd like to propose this solution to issues with the install script on ARM64 systems.
The default behaviour is to block the installation on ARM64 and printing out an error message which points out that official binaries are not available. But if the user specifies the `-y`-argument and agrees to use third party builds they can use the install script to deploy the binaries built by @LukeChannings to their system.

_PS: I'm using these builds for over a year without any trouble, that's why I think its a good idea to make them available official as an alternative._

this fixes #199

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -12,17 +12,25 @@ fi
 if [ "$OS" = "Windows_NT" ]; then
 	target="x86_64-pc-windows-msvc"
 else
+	repository="denoland/deno"
 	case $(uname -sm) in
 	"Darwin x86_64") target="x86_64-apple-darwin" ;;
 	"Darwin arm64") target="aarch64-apple-darwin" ;;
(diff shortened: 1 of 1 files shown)


ry:
For security we cannot have externally managed binaries installed from this script.

PR #212, opened 2022-06-16
fix: Use .bashrc instead of .bash_profile

nayeemrmn:
Closes https://github.com/denoland/vscode_deno/issues/392.

`.bash_profile` is under some circumstances not loaded by VSCode's extension host, making it fail to find `deno` if that's where the env setup is. `.bashrc` is more common.


bartlomieju (approved):
Thanks

PR #216, opened 2022-06-21
Remove Expand-Archive workaround

### LINKED ISSUE #214: Remove Expand-Archive workaround
https://github.com/denoland/deno_install/blob/40bb9a4d2ccfbc90ddf6855613682f53dbeaec8e/install.ps1#L40-L48

This workaround exists to support Windows 8.1 and older. Since 2016 (since PowerShell 5 was released with Windows 10), [`Expand-Archive`](https://docs.microsoft.com/de-de/previous-versions/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-5.0) is available.

See https://github.com/denoland/deno_install/issues/54 for why this workaround was added in the first place.

Windows 8.1 is already out of mainstream support and extended support will end beginning of next year (see 
https://docs.microsoft.com/en-us/lifecycle/products/windows-81) so I feel like it's safe to remove.

In the end, the 8 lines above should just be 1 line again:

```powershell
Expand-Archive $DenoZip -Destination $BinDir -Force
```

MarkTiedemann:
Fixes #214.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -37,15 +37,7 @@ if (!(Test-Path $BinDir)) {
 
 Invoke-WebRequest $DenoUri -OutFile $DenoZip -UseBasicParsing
 
-if (Get-Command Expand-Archive -ErrorAction SilentlyContinue) {
-  Expand-Archive $DenoZip -Destination $BinDir -Force
-} else {
-  if (Test-Path $DenoExe) {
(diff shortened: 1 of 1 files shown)


dsherret (approved):
LGTM. Thanks!

PR #219, opened 2022-06-21
Improve performance of PowerShell installer

### LINKED ISSUE #218: Improve performance of Powershell installer
On my machine, `install.ps1` is ~7x faster if I switch from [`Invoke-WebRequest`](https://github.com/denoland/deno_install/blob/master/install.ps1#L38) to `curl.exe` and from [`Expand-Archive`](https://github.com/denoland/deno_install/blob/master/install.ps1#L41) to `tar.exe`.

Before:
```powershell
PS C:\dev\deno_install> measure-command { .\install.ps1 } | select -expand totalseconds
13,1810087
```
After:
```powershell
PS C:\dev\deno_install> measure-command { .\install.ps1 } | select -expand totalseconds
1,8624925
```

This is similar to what's mentioned in #154:

> As brought up in https://github.com/denoland/deno_install/issues/153, since early 2018, `curl` and `tar` are available on Windows 10 by default. Thus, on a modern Windows 10 installation, `curl` could be used to download files and `tar` could be used to extract a `.zip` file.

The new implementation would look like this:

```powershell
curl.exe -Lo $DenoZip $DenoUri

tar.exe xf $DenoZip -C $BinDir
```

If this proposal would be accepted, https://github.com/denoland/deno_install/pull/216 could be closed.

MarkTiedemann:
Fixes #218.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -35,17 +35,9 @@ if (!(Test-Path $BinDir)) {
   New-Item $BinDir -ItemType Directory | Out-Null
 }
 
-Invoke-WebRequest $DenoUri -OutFile $DenoZip -UseBasicParsing
+curl.exe -Lo $DenoZip $DenoUri
 
-if (Get-Command Expand-Archive -ErrorAction SilentlyContinue) {
-  Expand-Archive $DenoZip -Destination $BinDir -Force
-} else {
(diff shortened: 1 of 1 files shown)


dsherret (approved):
Thanks! I've also noticed it was super slow.

PR #261, opened 2023-04-21
Make install script provide tips that are relevant to shell used by the user

erelinz:
This PR adds two statements to the install script, the statements check if your using a different shell as your default (other than the shell used to run the install script) and would recommend that you add deno to the path based on that shell default config files.

cc: @MarkTiedemann

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -42,17 +42,35 @@ unzip -d "$bin_dir" -o "$exe.zip"
 chmod +x "$exe"
 rm "$exe.zip"
 
+if [ "$(uname)" = "Darwin" ]; then
+	shell_name=$(dscl . -read ~/ UserShell | awk '{print $2}') # get the default shell for the user
+	shell_name=$(basename "$shell_name") # דet the shell name
+else
+	shell_name=$(getent passwd "$(id -un)" | cut -d: -f7) # get the default shell for the user
+	shell_name=$(basename "$shell_name") # דet the shell name
+fi
+
+
 echo "Deno was installed successfully to $exe"
 if command -v deno >/dev/null; then
 	echo "Run 'deno --help' to get started"
 else
-	case $SHELL in
-	/bin/zsh) shell_profile=".zshrc" ;;
-	*) shell_profile=".bashrc" ;;
+	case $shell_name in
+	zsh) shell_profile=".zshrc" ;;
+	bash) shell_profile=".bashrc" ;;
+	fish) shell_profile=".config/fish/config.fish" ;;
+	*) shell_profile=".profile" ;;
 	esac
+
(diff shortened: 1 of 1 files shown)


MarkTiedemann:
@erelinz I'm not a maintainer. Just a contributor. Here's my opinion:

While your changes may be technically correct, I'm not sure the added complexity is warranted.

Consider the following lines you added:

```
dscl . -read ~/ UserShell | awk '{print $2}'
```

```
getent passwd "$(id -un)" | cut -d: -f7
```

The installer script should be kept simple - so simple that you can read it in one go and be sure that it's not gonna harm your system. I'm not sure that's the case with these two commands, especially with the second one referring to `passwd`. I don't think these are very common. So it won't be simple for most users to understand them.

Therefore, I'd rather not add these changes.

erelinz:
Sorry @MarkTiedemann I saw you made lots of contributions and wrongly assumed. Please ignore.
> @erelinz I'm not a maintainer. Just a contributor. Here's my opinion:
> 
> While your changes may be technically correct, I'm not sure the added complexity is warranted.
> 
> Consider the following lines you added:
> 
> ```
> dscl . -read ~/ UserShell | awk '{print $2}'
> ```
> 
> ```
> getent passwd "$(id -un)" | cut -d: -f7
> ```
> 
> The installer script should be kept simple - so simple that you can read it in one go and be sure that it's not gonna harm your system. I'm not sure that's the case with these two commands, especially with the second one referring to `passwd`. I don't think these are very common. So it won't be simple for most users to understand them.

I think the your logic is fair but not the conclusion. The purpose of this PR is to make it easier and simple for users to install deno, regardless of their OS and shell of choice.

Users should be able to understand what the script does if they chose to review it. If they are unsure, they can paste it into chatgpt or https://explainshell.com/ and they'll be be satisfied with the answer.

The script uses command line utilities that are installed in many distributions, so no additional installation of packages would required for using those commands. We could also add more comments explaining why we want to tailor the output to the user's OS and shell.


I've tested it on macOS other linux distributions:
```
docker run -it --name alpine alpine sh -c "apk update && apk add curl && curl -fsSL https://raw.githubusercontent.com/erelinz/deno_install/fish_aware_tip/install.sh |sh"
docker run -it --name ubuntu ubuntu bash -c "apt update && apt -y install curl unzip && curl -fsSL https://raw.githubusercontent.com/erelinz/deno_install/fish_aware_tip/install.sh |sh"
```
Similarly for macOS, `dscl` and [`awk`](https://developer.apple.com/library/archive/documentation/OpenSource/Conceptual/ShellScripting/Introduction/Introduction.html#//apple_ref/doc/uid/TP40004268-TP40003516-SW1) are available by default.

PR #262, opened 2023-04-30
Add `wget` fallback support for minimal OS's

### LINKED ISSUE #256: Permission denied during install
When I try to install using
```
curl -fsSL https://deno.land/install.sh | sh
```
I get
```
Warning: Failed to open the file /home/jesper/.deno/bin/deno.zip: Permission  
Warning: denied
                                                                           0.0%curl: (23) Failure writing output to destination
```

jeff-hykin:
This should also fix  #256. This PR has minimal edits to use wget as a fallback of curl, which is useful for Docker instances and other minimal OS's

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,7 +5,7 @@
 set -e
 
 if ! command -v unzip >/dev/null; then
-	echo "Error: unzip is required to install Deno (see: https://github.com/denoland/deno_install#unzip-is-required)." 1>&2
+	echo "Error: unzip is required to install Deno (see: https://github.com/denoland/deno_install#unzip-is-required )." 1>&2
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


MarkTiedemann:
At the top of the installer script, it says "Keep this script simple". I'm not sure this change is simple since it's complicating the one-liners for most users.

If you don't have curl installed in your minimal system, then you probably also know how to make the necessary adjustments. So I'd prefer not to add this change.

jeff-hykin:
my understanding was that the simplicity was so that other people could pretty easily verify the code wasn't malicious or vulnerable, which is definitely something I'm onboard with. I didn't think the complexity of the one liner fell into that, so I optimized it for minimal characters.

would this be considered more simple? 
```
sh <<<"$(curl -fsSL https://deno.land/install.sh || wget -qO- https://deno.land/install.sh)"
```

In terms of script modification, do you know of a more simple way of addressing #256?

I'm working with undergrad students who install the base version of Ubuntu which doesn't have curl, so they're not exactly skilled users.

MarkTiedemann:
I think #256 is not really a deno_install issue, but rather a Ubuntu specific issue with some curl installations.

Then again, I'm not a maintainer, just a contributor, so it's not my call.

mmastrac:
I think that we can add `wget` support to install.sh, but we should have separate `wget` instructions as a fallback rather than putting both `cURL` and `wget` into a single command.

Here's what I suggest we do:

1. Revert the README changes and replace them with a "fallback" section (something like "if you don't have curl...")
2. Limit the install.sh changes to just this:

```
curl --fail --location --progress-bar --output "$exe.zip" "$deno_uri" || \
	wget --output-document="$exe.zip" "$deno_uri" || \
	echo "When installing deno, I looked for the 'curl' and for 'wget' commands but I didn't see either of them. Please install one of them, otherwise I have no way to install Deno" 
```

ghost (approved):

ghost (approved):

ghost (approved):

PR #281, opened 2024-02-07
Deno ARM install support

sant123:
See https://github.com/denoland/deno/pull/22298

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -9,33 +9,21 @@ if ! command -v unzip >/dev/null; then
 	exit 1
 fi
 
-function version { echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'; }
-
-if [ $# -gt 0 ]; then
-    deno_version=$1
-fi
-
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -9,24 +9,33 @@ if ! command -v unzip >/dev/null; then
...
+    deno_version=$1
+fi
+
 if [ "$OS" = "Windows_NT" ]; then
 	target="x86_64-pc-windows-msvc"
 else
 	case $(uname -sm) in
 	"Darwin x86_64") target="x86_64-apple-darwin" ;;
 	"Darwin arm64") target="aarch64-apple-darwin" ;;
 	"Linux aarch64")
```
mmastrac, line 24: I don't think we need to worry about a version test here -- we can just enable support.

sant123: Sounds good to me 🤝

littledivy (approved):
LGTM

mmastrac (approved):
LGTM!

Bostt (commented):
LGTM

OBITOONDEADO1MZ (approved):

PR #284, opened 2024-05-28
Add `install_global.ps1` for global system install

Mqxx:
This pull request adds another Powershell script for the global installation of Deno. In this script the global system PATH variable for Deno is set instead of the user PATH variable and the default installation directory is set to `Env:ProgramFiles`.

### DIFF
--- a/install_global.ps1
+++ b/install_global.ps1
@@ -1,0 +1,46 @@
+$ErrorActionPreference = "Stop"
+
+if ($v) {
+  $Version = "v${v}"
+}
+if ($Args.Length -eq 1) {
+  $Version = $Args.Get(0)
+}
+
+$DenoInstall = $env:DENO_INSTALL
+$BinDir = if ($DenoInstall) {
+  "${DenoInstall}\bin"
+} else {
+  "${Env:ProgramFiles}\deno\bin"
+}
+
+$DenoZip = "$BinDir\deno.zip"
+$DenoExe = "$BinDir\deno.exe"
+$Target = "x86_64-pc-windows-msvc"
+
+$DownloadUrl = if (!$Version) {
+  "https://github.com/denoland/deno/releases/latest/download/deno-${Target}.zip"
+} else {
+  "https://github.com/denoland/deno/releases/download/${Version}/deno-${Target}.zip"
+}
+
+if (!(Test-Path $BinDir)) {
+  New-Item $BinDir -ItemType Directory | Out-Null
+}
+
+curl.exe -Lo $DenoZip $DownloadUrl
+
+tar.exe xf $DenoZip -C $BinDir
+
+Remove-Item $DenoZip
+
+$Machine = [System.EnvironmentVariableTarget]::Machine
+$Path = [System.Environment]::GetEnvironmentVariable("Path", $Machine)
+if (!(";${Path};".ToLower() -like "*;${BinDir};*".ToLower())) {
+  [System.Environment]::SetEnvironmentVariable("Path", "${Path};${BinDir}", $Machine)
+  $Env:Path += ";${BinDir}"
+}
+
+Write-Output "Deno was installed successfully to ${DenoExe}"
+Write-Output "Run "deno --help" to get started"
+Write-Output "Stuck? Join our Discord https://discord.gg/deno"


install_global.ps1
```diff
@@ -0,0 +1,46 @@
...
+}
+
+curl.exe -Lo $DenoZip $DownloadUrl
+
+tar.exe xf $DenoZip -C $BinDir
+
+Remove-Item $DenoZip
+
+$Machine = [System.EnvironmentVariableTarget]::Machine
+$Path = [System.Environment]::GetEnvironmentVariable("Path", $Machine)
```
dsherret, line 38: Maybe this should be an argument to the original script instead of a completely new script?

Mqxx: Okay, I didn't want to change the original script to avoid breaking changes.

Mqxx: And yes, you are right, the environment variable is the only difference between the two scripts.

Mqxx: > this should be an argument to the original script

I'm also not sure how to “parameterize” this...

dsherret: Probably it should just be a flag to the script. It shouldn't be breaking.

dsherret:
@joaka1987 can you please stop spamming?

install_global.ps1
Mqxx: Okay, I will update this as soon as possible 👍

Mqxx: All done

Mqxx:
The only thing I am not so sure is how to call the script from the URL using parameters...

Mqxx:
Hey @dsherret, can this be merged?

install.ps1
```diff
@@ -6,14 +6,15 @@ $ErrorActionPreference = 'Stop'
 
 if ($v) {
   $Version = "v${v}"
-}
-if ($Args.Length -eq 1) {
-  $Version = $Args.Get(0)
+} elseif ($Args.Length -eq 1) {
```
dsherret, line 9: I don't think this should be an elseif because now `$v` being set takes priority, but it should be the other way around.

README.md
```diff
@@ -18,6 +18,14 @@ curl -fsSL https://deno.land/install.sh | sh
 irm https://deno.land/install.ps1 | iex
 ```
 
+## Install Globally
+
+**With PowerShell:**
+
+```powershell
+$g; irm https://deno.land/install.ps1 | iex
+```
```
dsherret, line 27: Can you move this below "Install via Package Manager"?

install.ps1
```diff
@@ -38,10 +39,15 @@ tar.exe xf $DenoZip -C $BinDir
 
 Remove-Item $DenoZip
 
-$User = [System.EnvironmentVariableTarget]::User
-$Path = [System.Environment]::GetEnvironmentVariable('Path', $User)
+$EnvTarget = if (Test-Path variable:\g) {
```
dsherret, line 42: Maybe instead of `$g` we should use something like `$denoGlobalInstall`/`$denoInstallGlobally` or something long and specific. That way it doesn't accidentally install globally for someone who happens to have `$g` variable and then they get confused. With `$v` it's not so bad if it accidentally happens.

dsherret (changes requested):
Thanks! Almost there.

install.ps1
Mqxx: Okay got it 👍🏻

README.md
Mqxx: I will do so.

install.ps1
Mqxx: Okay, yes I agree.

Mqxx:
I will make these changes asap.

Mqxx:
@dsherret Done 👍

dsherret:
Hmmm... not sure why the mac ci is failing. Must be unrelated.

Mqxx:
Can you re-run it?

Mqxx:
Hey @dsherret, can you re-run it?

Mqxx:
@dsherret can this be merged?

Mqxx:
For some reason the pipeline fails. I am not sure why.

PR #295, opened 2024-09-18
feat: Add a script to set up the PATH and install shell completions

nathanwhit:
Ref https://github.com/denoland/deno/issues/24157

Written in typescript for maintainability and ease of contribution (I am not good at bash scripts, and I think most people are more comfortable with TS than bash). The downside is that we only support this for deno 1.42 and above. I think this is fine because most people installing deno interactively are likely to be installing new deno versions.

The actual PATH setup logic and general shape is largely adapted from rustup.

Here's what it looks like currently:

https://github.com/user-attachments/assets/1fb4200f-538b-4237-a90c-55849e870a3a

Some features:
- Supports sh, bash, zsh, and fish
- Set up PATH vars for supported shells
- Optional completion setup for supported shells
- Backs up rc files in case you want to revert

Things I am unsure about:
- Whether to prompt for writing to outside of the install dir
  - currently I only allow-write for the install dir (~/.deno) 
  - I thought it was a nice opportunity to show of the permission system, and so users know what we're modifying, but it may be more noise than it's worth
- Whether bash completions work (we recommend installing to /usr/local/etc, but that requires root)


bartlomieju:
> Whether to prompt for writing to outside of the install dir
currently I only allow-write for the install dir (~/.deno)
I thought it was a nice opportunity to show of the permission system, and so users know what we're modifying, but it may be more noise than it's worth

That seems undesirable IMO - running installation script shouldn't prompt at all - we should add entry to `.bashrc` by default and print information about it. Power users who don't want that we'll be able to remove it.

LICENSE
```diff
@@ -0,0 +1,9 @@
+MIT License
```
littledivy, line 1: lol

littledivy (approved):

littledivy:
> That seems undesirable IMO - running installation script shouldn't prompt at all

I agree. The shell script writes out `deno` binary to file system anyways so asking for permission doesn't make sense.

lucacasonato:
It would be nice if we bundled the code here into a single JS file and ran that. Pulling in dax in the installer is kinda annoying.

install.sh
```diff
@@ -45,16 +45,18 @@ chmod +x "$exe"
...
 
 echo "Deno was installed successfully to $exe"
+if $exe eval 'const [major, minor] = Deno.version.deno.split("."); if (major < 2 && minor < 42) Deno.exit(1)'; then
+
+	allow_sys_perm="--allow-sys=homedir"
+
+	if ! echo | $exe run --allow-sys=homedir - >/dev/null 2>&1; then
+		allow_sys_perm="--allow-sys"
+	fi
+	$exe run $allow_sys_perm --no-check --allow-run="zsh,$exe" --allow-read --allow-env --allow-write="$deno_install" ./shell-setup/src/main.ts "$deno_install"
```
kt3k, line 55: Will this part (`./shell-setup/src/main.ts`) be updated to jsr url later?

nit: Is `--no-check` flag necessary?

nathanwhit:
Okay, I updated this to bundle into one entrypoint, and also removed the dependency on dax (I was just using it for prompts, and it's overkill. Instead I just adapted some of the code from dax, replacing the rust WASM dependencies with js, and published a little library https://jsr.io/@nathanwhit/promptly). That brings the bundled size down to 20KB.

Also updated to avoid permission prompts, as suggested.

---

All that's left is to publish to JSR, and then this should be good to go.

install.sh
nathanwhit: Yup, was planning on updating it to jsr (just did it now). Removed `--no-check` as well.

nathanwhit:
Published to JSR, this should be ready to go.

Just need to merge this PR and update `deno.land/install.sh`

install.sh
```diff
@@ -45,16 +45,18 @@ chmod +x "$exe"
...
 
 echo "Deno was installed successfully to $exe"
+if $exe eval 'const [major, minor] = Deno.version.deno.split("."); if (major < 2 && minor < 42) Deno.exit(1)'; then
+
+	allow_sys_perm="--allow-sys=homedir"
+
+	if ! echo | $exe run --allow-sys=homedir - >/dev/null 2>&1; then
+		allow_sys_perm="--allow-sys"
+	fi
+	$exe run $allow_sys_perm --allow-run="zsh,$exe" --allow-read --allow-env --allow-write jsr:@deno/installer-shell-setup/bundled "$deno_install"
```
bartlomieju, line 55: Maybe just `-A` to be sure?

shell-setup/src/environment.ts
```diff
@@ -0,0 +1,55 @@
...
+    return await Deno.stat(path);
+  } catch (error) {
+    if (
+      error instanceof Deno.errors.NotFound ||
+      (error instanceof Deno.errors.PermissionDenied &&
+        (await Deno.permissions.query({ name: "read", path })).state ==
+          "granted")
+    ) {
+      return;
+    }
```
bartlomieju, line 19: Running with `-A` would make this code not required?

shell-setup/src/main.ts
```diff
@@ -0,0 +1,330 @@
...
+    await addToPath(availableShells, installDir, backups);
+  }
+
+  const shellsWithCompletion = availableShells.filter((s) =>
+    s.supportsCompletion !== false
+  );
+  const selected = await multiSelect(
+    {
+      message: `Set up completions?`,
+      options: shellsWithCompletion.map((s) => {
```
bartlomieju, line 287: So this is an interactive prompt - are we sure we want to prompt user and require more interaction? I'm fine either way.

bartlomieju (approved):

shell-setup/src/environment.ts
nathanwhit: Yeah that's true. I'd prefer to leave it in though, so that this script works regardless of the permission flags

install.sh
nathanwhit: Done

shell-setup/src/main.ts
nathanwhit: For now I'd prefer to have it opt-in, since setting up completions is more invasive that adding to path, and more dependent on your shell setup.

We can change it in the future if it's too annoying

bartlomieju (approved):

PR #305, opened 2024-10-01
chore: fix publishing

nathanwhit:
There's an implicit `@types/node` dependency when publishing, since we use node builtin modules. Combined with a frozen lockfile this caused publishing to fail.

To catch this in the future, I added `deno publish --dry-run` to CI.


littledivy (approved):
LGTM

Issue #311, opened 2024-11-09
installing deno on alpine leads to deno: not found


BlackScorp:
Hi all, i want to create a frontend service in a custom docker container but after installing deno, it is not found on cmd but the binary is there. maybe the binary is not compatible with alpine linux?

I use following Docker file

```
ARG ALPINE_VERSION=${ALPINE_VERSION:-latest}
FROM alpine:$ALPINE_VERSION


RUN apk -U upgrade
RUN apk add --no-cache \
    curl

RUN export DENO_INSTALL=/usr/local && curl -fsSL https://deno.land/install.sh |  sh

CMD ["tail","-f","/dev/null"]
```
the output of the installation is following

```
######################################################################## 100.0%
Archive:  /usr/local/bin/deno.zip
  inflating: deno
Deno was installed successfully to /usr/local/bin/deno
sh: /usr/local/bin/deno: not found
Run 'deno --help' to get started

Stuck? Join our Discord https://discord.gg/deno
```

you can already see here the not found error.

so i checked /usr/local/bin

```
total 138384
drwxr-xr-x    1 root     root          4096 Nov  9 21:22 .
drwxr-xr-x    1 root     root          4096 Sep  6 11:34 ..
-rwxr-xr-x    1 root     root     141686664 Nov  9 21:22 deno
```
the file is there and is executable

echo $PATH shows following

```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

so executable is there but is for some reason it is not found. i cannot use alpines apk add deno because it is in version 1.4 and i would like to use the latest one.

maybe someone has an idea?

jabez007:
I'm running into the same issue, and I think it's something to do with this section of the `install.sh` basically skipping `run_shell_setup` which is where the PATH is updated with the location of deno
```
# If stdout is a terminal, see if we can run shell setup script (which includes interactive prompts)
if [ -z "$CI" ] && [ -t 1 ] && $exe eval 'const [major, minor] = Deno.version.deno.split("."); if (major < 2 && minor < 42) Deno.exit(1)'; then
	if [ -t 0 ]; then
		run_shell_setup "$@"
	else
		# This script is probably running piped into sh, so we don't have direct access to stdin.
		# Instead, explicitly connect /dev/tty to stdin
		run_shell_setup "$@" </dev/tty
	fi
fi
```

jabez007:
this should be fixed now by passing `-y` to the install script

for example
```
[...]

PR #312, opened 2024-11-15
Allow run_shell_setup during Docker build

jabez007:
In reference to this issue
https://github.com/denoland/deno_install/issues/311

The idea is to use an ARG in the Dockerfile as a flag to indicate that we are in a build so that the `install.sh` will still execute `run_shell_setup`, and since we are in a Docker build pass the option to `run_shell_setup` so that the prompts are just answered with their defaults.

an example Dockerfile then would look something like
```
FROM debian:stable-slim

# Define a build argument to indicate the Docker build context
ARG DOCKER_BUILD=true

RUN apt-get update && \
  apt-get install -y \
  bash \
  curl \
  unzip

ENV SHELL="/bin/bash"

#RUN curl -fsSL https://deno.land/install.sh | sh
RUN curl -fsSL https://raw.githubusercontent.com/jabez007/deno_install/refs/heads/fix-docker/install.sh | sh

CMD ["/bin/bash"]
```


nathanwhit:
Hello, thanks for the PR!

I think instead of using an ARG and special casing docker builds, we should check if the user passed in `-y` or `--yes`, and if so run the shell setup regardless of whether it's an interactive terminal.

The steps for that would be roughly:

1. Update the arg parsing [here](https://github.com/denoland/deno_install/blob/b4dad3254e44fbad38f7a4859e7840342cdba2b9/install.sh#L49) to check for -y/--yes, and if it's present set some var (`should_run_shell_setup` or something)
2. Update the condition to be `if { [ -z "$CI" ] && [ -t 1 ]; } || $should_run_shell_setup;`

That way you could install deno in a dockerfile by just passing `-y`, so it would be consistent across environments instead of being dockerfile specific

nathanwhit (approved):
LGTM, thanks!

PR #314, opened 2024-11-18
fix: correct install paths for bash completions

uncomfyhalomacro:
Closes: https://github.com/denoland/deno/issues/26855


nathanwhit (approved):
LGTM, thanks for the PR!

PR #315, opened 2024-11-21
fix: verify script now runs correctly

uncomfyhalomacro:
previous command does not work as I receive this output

```
shasum: /proc/self/fd/11: no properly formatted SHA checksum lines found
```


nathanwhit (approved):
LGTM, thanks!

PR #337, opened 2025-12-12
fix newline bug, install dx alias, add unit testing, switch to deno bundle

### LINKED ISSUE #335: Deno appends line to `.zshrc`/`.bashrc` without trailing newline
When installing via the `install.sh` script, it appends a line to the shell rc files, but that line doesn’t have a trailing newline.

This is a problem when installing multiple tools with similar install procedures e.g. in a Dockerfile, since the next installation will append to the end of the line rather than the next line.

```
❯ curl -fsSL https://deno.land/install.sh | sh -s -- -y
. . .
❯ tail -1 ~/.zshrc ; echo EOF
. "/home/daniel/.deno/env"EOF
❯
```

Correct:

```
❯ tail -1 ~/.zshrc ; echo EOF
. "/home/daniel/.deno/env"
EOF
❯
```

### LINKED ISSUE #336: Install `dx` alias by default
2.6 introduced `dx` cli util. https://deno.com/blog/v2.6#run-package-binaries-with-dx

Maybe we should install that alias by default with the install script like node installs `npx` by default

nathanwhit:
Fixes #335.
Fixes #336.


bartlomieju (approved):
Thanks!

PR #321, opened 2025-01-29
install: allow `busybox unzip`

erg:
according to issue #172, `busybox unzip` is an alternative to installing unzip or 7z.

busybox is installed by default on Ubuntu 24.04.1 LTS but unzip/7z are not.

note: busybox is not installed on macOS 15.3 by default but unzip is.

PR #338, opened 2025-12-28
Allow various commands to verify and extract the archive

### LINKED ISSUE #172: Allow (bsd)tar to unzip
Currently, we require `unzip` to be installed for extracing the Zip archive (see https://github.com/denoland/deno_install/blob/master/install.sh#L7-L10, https://github.com/denoland/deno_install#unzip-is-required).

Contrary to `gnutar`, `bsdtar` (a `tar` implementation by [libarchive](https://github.com/libarchive/libarchive)) is able to extract `.zip` files, too. It's the default `tar` on FreeBSD, NetBSD, MacOS and Windows (see https://github.com/libarchive/libarchive/wiki/LibarchiveUsers).

So if `unzip` is unavailable, we could check for `(bsd)tar`, and only if both are unavailable, fail with an error message.

I'm using a similar technique in [deno.mk](https://github.com/MarkTiedemann/deno.mk) (see https://github.com/MarkTiedemann/deno.mk/blob/master/deno.mk#L64-L68).

### LINKED ISSUE #260: install.sh does not work on Synology DSM
I can't use this install script to install deno on a Synology DSM Instance (with x86 processor), 
since DSM on Synology does not ship with 'unzip' anymore. 
Re-installing it is not trivial, since there is no package manager on DSM. 

I'm currently working on a small pr to have an install script with support for 7z.

### LINKED ISSUE #266: Installing on 32bit linux should throw an error

### LINKED ISSUE #322: Install location of the deno-CLI  is not XDG compliant
Hi, im new to deno and the first thing i noticed is that it installs under `~/.deno/bin` 🤔 
This needs a modification of the ~$PATH` variable ofcourse, but why?

> [XDG Base Directory Specification - 3 Environment variables](https://specifications.freedesktop.org/basedir-spec/latest/#variables)
> User-specific executable files may be stored in $HOME/.local/bin.
> Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place. 

The above is already done in most distro's, `Kubuntu` for sure as i use that at moment...

So i suggest to install the CLI as `~/.local/bin/deno`, which doesn't need any change to the `$PATH`.

The below script used to install `deno` should be changed for this: https://deno.com/
```console
curl -fsSL https://deno.land/install.sh | sh
```

### LINKED ISSUE #325: Basically any error hides "Stuck? Join us on discord" message
My initial issue was a VPN (mullvad) causing the following:

```sh
> curl -fsSL https://deno.land/install.sh | sh
curl: (3) URL using bad/illegal format or missing URL
```
(no other output)

However, after reading the installer, I realized basically the only time "Stuck? Join us on discord" gets printed is if there are no errors at all, e.g. the only time the debugging message is printed is when the user is least likely to be debugging.

## Easy reproduction

- Download the installer script and add "_" (or any char) to `https://dl.deno.land/release-latest.txt` and it will effectively simulate a vpn blockage.
- Alternatively, also try appending something to the end of the deno_uri to see how it causes no help messages to be printed

tcely:
Replaces:

- #302 
- #321 
- #326 
- #333 

Fixes:
- fix #172 
- fix #266 
- fix #322 
- fix #325 

Closes:
- close #260


teetee4482-rgb (approved):
- [ ]

PR #351, opened 2026-07-15
feat: install native Windows ARM64 Deno

### LINKED ISSUE #347: install.ps1 does not detect Windows ARM and always installs x86_64 version
The PowerShell installation script does not detect the system architecture on Windows ARM devices. When running the install command on a Windows 11 ARM laptop:

```powershell
irm https://deno.land/install.ps1 | iex
```

The script always downloads the x86_64 version (`x86_64-pc-windows-msvc`) regardless of the system's actual architecture, because the target is hardcoded on line 19 of the script:

```powershell
 = 'x86_64-pc-windows-msvc'
```

While Deno does provide ARM64 builds for Windows (as noted in their 2.7 release: "Windows ARM" support), the install script does not utilize them. Users on Windows ARM devices get the x86_64 binary that runs via emulation instead of the native ARM64 version.

**Expected behavior:**
The script should detect the system architecture and download the appropriate version:
- `x86_64-pc-windows-msvc` for x86_64 systems
- `aarch64-pc-windows-msvc` for ARM64 systems

**Environment:**
- Windows 11 ARM (e.g., Surface Pro X, Samsung Galaxy Book, etc.)
- PowerShell

bartlomieju:
## Summary

Makes the official Deno installers select the **native** Windows ARM64 build. Previously both `install.ps1` and `install.sh` hardcoded `x86_64-pc-windows-msvc`, so a Windows ARM64 machine downloaded the x64 binary and ran it under emulation.

This is §2 of denoland/deno#36045, and closes #347.

After installing on Windows ARM64, `deno eval 'console.log(Deno.build.target)'` now reports `aarch64-pc-windows-msvc`.

### Native architecture detection
- **`install.ps1`**: `Get-DenoTarget` calls [`IsWow64Process2`](https://learn.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2) and uses its *native-machine* result — robust even when x64 PowerShell (including Windows PowerShell 5.1 on .NET Framework) runs under emulation. `RuntimeInformation.OSArchitecture` is a guarded fallback for hosts where the P/Invoke is unavailable.
- **`install.sh`** (MSYS/Git-Bash path): reads `PROCESSOR_ARCHITEW6432` (set to the native arch under WOW64 emulation) falling back to `PROCESSOR_ARCHITECTURE`.
- Native ARM64 → `aarch64-pc-windows-msvc`, native AMD64 → `x86_64-pc-windows-msvc`. **x64 Windows behavior is unchanged.**

### Version guard
Native Windows ARM64 artifacts are first published in [v2.6.8](https://github.com/denoland/deno/releases/tag/v2.6.8). Requesting an older version on ARM64 now **fails with a clear error** rather than silently installing the emulated x64 build.

### Checksum verification (`install.ps1`)
Downloads and verifies `deno-<target>.zip.sha256sum` before extraction, removing temp files on any failure. The parser accepts both canonical `<hash>  <file>` and legacy PowerShell `Format-List` checksum files. Verification is **fatal on ARM64** (a checksum is always expected there); on x64 it is applied when a checksum is present, so pre-2.6.7 releases that predate published checksums still install.

### CI
New `test-windows-arm` job on `windows-11-arm` runs `install_test_arm.ps1`:
1. Native ARM64 install (latest) asserts `Deno.build.target`.
2. A pinned ARM64-capable version (`2.6.8`) installs native.
3. A pre-2.6.8 request (`2.6.7`) is rejected.
4. An **emulated x86 PowerShell** (SysWOW64) install still selects the native ARM64 build — exercising the `IsWow64Process2` path (skipped if that host is absent on the image).

Legacy v1.x installs remain on the x64 runner because those releases have no ARM64 artifacts.

## Verified locally
`shellcheck -s sh *.sh`, `shfmt -d .`, `deno fmt --check`, `deno lint`, and `shasum -a 256 -c SHA256SUM` all pass. `SHA256SUM` was regenerated for the updated scripts.

## Not in this PR
- `install.sh` checksum verification for all platforms (Linux/macOS have none today) — separate concern.
- Publishing a new tag and updating the `deno.com/install.ps1` redirect — release infra, tracked in #36045.


bartlomieju:
Review notes. Overall solid — `install.ps1` uses a robust `IsWow64Process2` probe and the version-guard arithmetic is correct for all real three-component versions. A few things worth addressing, most-severe first.

### 1. `install.sh` native ARM64 detection likely fails for 64-bit x64 Git Bash (the common case)

```sh
native_arch="${PROCESSOR_ARCHITEW6432:-$PROCESSOR_ARCHITECTURE}"
```

`PROCESSOR_ARCHITEW6432` is documented as set only inside a WOW64 process — a *32-bit* process on 64-bit Windows. Git for Windows ships as a 64-bit x64 build, which on ARM64 runs under x64 emulation and is **not** WOW64, so that variable is unset and `PROCESSOR_ARCHITECTURE` reports `AMD64`. Nothing matches `ARM64`, `target` falls back to `x86_64-pc-windows-msvc`, and the ARM64 user silently gets the emulated x64 build via the `curl … | sh` path. The `ARCHITEW6432` branch only ever fires for a rare 32-bit x86 shell.

Note the asymmetry: `install.ps1` deliberately avoids env vars in favor of `IsWow64Process2` ("robust source of truth"), but `install.sh` relies solely on the WOW64-only env var. Worth confirming with a real `echo $PROCESSOR_ARCHITEW6432` in 64-bit Git Bash on a physical ARM64 device — if it's empty there (as the WOW64 docs imply), the shell installer needs a different heuristic.

Refs: [WOW64 implementation details](https://learn.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details), [Detect x86_64 process on ARM64](https://learn.microsoft.com/en-us/answers/questions/449019/detect-x86-64-process-on-arm64).

### 2. `install.sh`'s ARM64 path has zero CI coverage

`test-windows-arm` runs only `install_test_arm.ps1`. The `tests shell` step runs `install_test.sh` exclusively on the x64 runners, where the Windows path always resolves to `x86_64`. Nothing exercises `install.sh` with `target=aarch64-pc-windows-msvc`, which is exactly why a detection bug like #1 would ship undetected. Consider a Git-Bash job on the `windows-11-arm` runner that asserts the resolved target.

### 3. `install_test_arm.ps1` test #2 never asserts the result

The pinned-`2.6.8` step omits `Assert-NativeTarget`, unlike tests #1 and #4, so it only checks that the install doesn't error — an arm64 regression on the explicit-version path would pass silently.

### 4. `install_test_arm.ps1` test #3 is coupled to `Write-Error`, not `exit 1`

`exit 1` inside a `.\install.ps1` invocation returns to the caller without throwing, so the `try/catch` wouldn't catch it. Test #3 passes only because the preceding `Write-Error` under `$ErrorActionPreference='Stop'` throws a terminating error that crosses the script boundary (making every `exit 1` in the script effectively dead code). If the arm64 guard later moves above the `$ErrorActionPreference='Stop'` line, or `Write-Error` becomes a plain `exit 1`, the throw disappears and `$rejected` stays `$false`. An explicit `$LASTEXITCODE`-based assertion would be more durable.

---

Investigated and dropped: a concern that the x64 "no published checksum → warn and continue" fallback is unreachable on pwsh 7.4+ (`$PSNativeCommandUseErrorActionPreference` + `Stop` making `curl --fail`'s non-zero exit throw). Refuted — the green `pwsh` CI run installs checksum-less v1.0.0/v1.1.0/v1.0.1 and passes, so the fallback is reached in practice.

bartlomieju:
Thanks — addressed in 345ce7a.

**1. `install.sh` x64-Git-Bash detection.** Correct, and the most important catch. `PROCESSOR_ARCHITEW6432` is WOW64-only (32-bit process), so 64-bit x64 Git Bash on ARM64 saw `AMD64` and fell back to x64. Now, when the env vars don't say `ARM64`, it reads the physical CPU identifier from `HKLM\HARDWARE\DESCRIPTION\System\CentralProcessor\0\Identifier` via `reg`, which x64/x86 emulation does not rewrite (`ARMv8…` → aarch64). `MSYS_NO_PATHCONV=1` keeps Git Bash from mangling the key/flag. I can't validate on a physical ARM64 device from here, so this leans on the new CI job below.

**2. No sh CI coverage.** Added `install_test_arm.sh` and wired it into the `windows-11-arm` job via `shell: bash`, so the shell installer's ARM64 path is now actually exercised (whichever way the runner's Git Bash resolves).

**3. Test #2 assertion.** This one is already covered — line 27 calls `Assert-NativeTarget "$Home\deno-arm64\bin\deno.exe"` after the pinned-`2.6.8` install, and the green run logged `OK: …\deno-arm64\bin\deno.exe reports aarch64-pc-windows-msvc`. No change needed, but flag me if you were looking at an earlier revision.

**4. Test #3 coupling to `Write-Error`.** Good durability point. Now checks `$LASTEXITCODE` in addition to the `catch`, so it holds even if the guard later becomes a plain `exit 1` or moves above `$ErrorActionPreference='Stop'`.

Also confirming your dropped concern: the x64 "no checksum → warn and continue" fallback is genuinely reached — the green `pwsh` run installs checksum-less v1.0.0/v1.0.1/v1.1.0.

